PR #12, opened 2023-08-12
Log level

maxdeichmann:


maxdeichmann:
@marcklingen i just built this in the aircraft. Log levels adjustable via flag. The `test_atexit` test only works because of adjusted log levels. New behavior: Log everything in debug mode, only log exceptions otherwise. I did not want to push this to main as i am on vacation now. Feel free to review and do so.

PR #18, opened 2023-08-16
Change in logger & fix for huggingface_hub

Dev-Khant:
Fixes : #16

PR #20, opened 2023-08-16
Test: Change in logger & fix for huggingface_hub (#18)

marcklingen:
* Change in logger & fix for huggingface_hub

* add test & enable langchain e2e tests by default


marcklingen:
@Dev-Khant thanks for your contribution

PR #35, opened 2023-08-23
Add end() to Generation and Span

Dev-Khant:


Dev-Khant:
@marcklingen @maxdeichmann Let me know of any required changes.

langfuse/client.py
```diff
@@ -138,8 +139,7 @@ def create_trace():
 
             def create_span():
                 try:
-                    new_body = body.copy(update={"id": new_span_id})
-                    new_body = body.copy(update={"trace_id": new_trace_id})
+                    new_body = body.copy(update={"id": new_span_id, "trace_id": new_trace_id})
```
Dev-Khant, line 142: Reason : 2 seperate spans were created from same trace because before this change when setting `trace_id` , `id` was set to None.

langfuse/client.py
```diff
@@ -179,8 +179,7 @@ def create_trace():
 
             def create_generation():
                 try:
-                    new_body = body.copy(update={"id": new_generation_id})
-                    new_body = body.copy(update={"trace_id": new_trace_id})
+                    new_body = body.copy(update={"id": new_generation_id, "trace_id": new_trace_id})
```
Dev-Khant, line 182: Reason : 2 seperate generations were created from same trace because before this change when setting `trace_id` , `id` was set to None.

langfuse/api/resources/span/types/update_span_request.py
```diff
@@ -12,6 +12,7 @@
 class UpdateSpanRequest(pydantic.BaseModel):
     span_id: str = pydantic.Field(alias="spanId")
     trace_id: typing.Optional[str] = pydantic.Field(alias="traceId")
+    name: typing.Optional[str]
```
maxdeichmann, line 15: Good catch! All files in langfuse/api are auto-generated by fern. Added a PR on the core project and will update this PR as well https://github.com/langfuse/langfuse/pull/101

By keeping the openapi docs in core in sync, we can also update the TS SDK accordingly.

langfuse/client.py
maxdeichmann: good catch!

langfuse/client.py
maxdeichmann: same here!

langfuse/client.py
```diff
@@ -190,8 +189,8 @@ def create_generation():
                     self.log.exception(e)
                     raise e
 
-            self.task_manager.add_task(new_generation_id, create_generation)
             self.task_manager.add_task(new_trace_id, create_trace)
+            self.task_manager.add_task(new_generation_id, create_generation)
```
maxdeichmann, line 193: Ordering does not matter, but more logical to have it this way

langfuse/client.py
```diff
@@ -358,7 +357,7 @@ def update(self, body: UpdateGeneration):
 
             def task():
                 try:
-                    new_body = body.copy(update={"generation_id": self.id})
+                    new_body = body.copy(update={"generation_id": self.id, "trace_id": self.trace_id})
```
maxdeichmann, line 360: Why this change? It does not destroy anything but did it not work without?

tests/test_sdk.py
```diff
@@ -424,3 +426,76 @@ def test_create_trace_with_id_and_generation():
...
+
+    assert len(langfuse.task_manager.result_mapping) == 3
+
+    trace_id = langfuse.get_trace_id()
+
+    trace = api_wrapper.get_trace(trace_id)
+
+    span = trace["observations"][0]
+    iso_format_end_time = end_time.astimezone(pytz.UTC).strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3] + "Z"
+    assert span["endTime"] == iso_format_end_time
```
maxdeichmann, line 469: Do you think these type of tests work reliably as they base on time? Maybe we should just check that the endtime is not null?

Dev-Khant: I thought on the platform time is shown in UTC time zone. But yes also checking if it's null or not is better. I'll make the change.

langfuse/client.py
Dev-Khant: Before this change it was not actually updating the generation because it couldn't find which trace it is.

langfuse/client.py
Dev-Khant: Yes

langfuse/api/resources/span/types/update_span_request.py
Dev-Khant: Alright

Dev-Khant:
@maxdeichmann I made the change to check for Null for endTime.

langfuse/client.py
```diff
@@ -401,6 +416,14 @@ def task():
         except Exception as e:
             self.log.exception(e)
 
+    def end(self):
+        try:
+            end_time = datetime.now()
+            self.update(UpdateGeneration(endTime=end_time))
```
maxdeichmann, line 422: I missed this one here: i think we need to return this function here so that end() returns a stateful client (also above). Otherwise good for me.

maxdeichmann (approved):

PR #67, opened 2023-09-08
fix llama.cpp model parameter

yigitbey:
This patch adds support for llama.cpp. Parameter name for model  is called `model_path` in llama.cpp, causing issue during trace generation


maxdeichmann:
@yigitbey thanks a lot for the contribution! Were you able to test this locally?

yigitbey:
yes and it works fine.
Screenshot 2023-09-08 at 14 58 20

marcklingen:
Thank you @yigitbey!

PR #100, opened 2023-09-20
Use langfuse keys from os

Dev-Khant:
Contributes to https://github.com/langfuse/langfuse/issues/232


langfuse/client.py
```diff
@@ -36,9 +36,6 @@ class Langfuse(object):
 
     def __init__(
         self,
-        public_key: str,
-        secret_key: str,
-        host: Optional[str] = None,
```
marcklingen, line 41: For backwards compatibility, I'd suggest to keep the constructor arguments (public, secret, host) and make them optional. This applies to the SDK and the langchain callback handler. 

If set, constructor arguments then take precedence over environment variable. Removing the argument leads to unexpected behavior for existing users

marcklingen (changes requested):
Thanks for your contribution on thi, see comment above regarding backwards compatibility

langfuse/client.py
Dev-Khant: Yes, it makes sense I have made the changes. And if access keys are not present, we are already showing a warning.

langfuse/callback.py
```diff
@@ -34,14 +35,17 @@ def __init__(
...
+
         # If we're provided a stateful trace client directly
         if statefulTraceClient:
             self.trace = Run(statefulTraceClient, None)
             self.runs = {}
 
         # Otherwise, initialize stateless using the provided keys
         elif public_key and secret_key:
-            self.langfuse = Langfuse(public_key, secret_key, host, debug=debug, release=release)
+            self.langfuse = Langfuse(debug=debug, release=release)
```
maxdeichmann, line 48: Should we set the keys here?

Dev-Khant: Yes makes more sense above we are accessing it so here we can keep it

marcklingen (changes requested):

langfuse/callback.py
```diff
@@ -31,6 +32,9 @@ def __init__(
         statefulClient: Optional[StatefulTraceClient | StatefulSpanClient] = None,
         release: Optional[str] = None,
     ) -> None:
+        public_key = public_key if public_key else os.environ.get("LF_PK")
```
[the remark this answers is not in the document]
Dev-Khant: Makes more sense, didn't see it

tests/api_wrapper.py
```diff
@@ -1,10 +1,13 @@
+import os
 import requests
 
 
 class LangfuseAPI:
-    def __init__(self, username=None, password=None, base_url="http://localhost:3000"):
-        self.auth = (username, password) if username and password else None
-        self.BASE_URL = base_url
+    def __init__(self, username=None, password=None, base_url=None):
```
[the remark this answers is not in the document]
marcklingen: Ok perfect, agree that we should not optimize for tests here

marcklingen (approved):

maxdeichmann (approved):

maxdeichmann:
Thanks for the contribution @Dev-Khant

PR #102, opened 2023-09-21
Show token usage and formatting of code

Dev-Khant:
Resolves https://github.com/langfuse/langfuse/issues/190

I think this is the necessary info for users. Let me know if we want to add more. Thanks :)


langfuse/callback.py
```diff
@@ -524,7 +538,9 @@ def on_llm_end(
                 )
 
                 self.runs[run_id].state = self.runs[run_id].state.update(
-                    UpdateGeneration(completion=extracted_response, end_time=datetime.now(), usage=llm_usage)
+                    UpdateGeneration(
+                        completion=extracted_response, end_time=datetime.now(), usage=llm_usage, metadata=token_metadata
```
maxdeichmann, line 542: Why do you want to put the token metadata on the metadata as well? We are already tracking that in `usage`.

Dev-Khant: I thought of giving all the information in metadata. But will remove it because in `usage` it's included.

Dev-Khant:
@maxdeichmann Here will we get the `headers` from the response because base class `LLMResult` does not contain headers? So we have to find a way to get headers in `on_llm_end`

langfuse/callback.py
maxdeichmann: Awesome

maxdeichmann:
> @maxdeichmann Here will we get the `headers` from the response because base class `LLMResult` does not contain headers? So we have to find a way to get headers in `on_llm_end`

Yes, i am not sure whether and if yes, how Langchain exposes these headers. Did you print out all the kwargs and the LLMResult? They might be in there.

Dev-Khant:
Here I think they are using `chatcompletion` method because in LLMResults they have no header info.
And all the info are only generated in  `chatcompletion`
image

Dev-Khant:
Here I think they are using `chatcompletion` method because in LLMResults they have no header info.
And all the info are only generated in  `chatcompletion`
image

maxdeichmann:
@Dev-Khant thanks a lot for the contribution. I hope it is fine for you if i close this as we are not able to get more header context from langchain unfortunately. Again, thanks a lot!

PR #103, opened 2023-09-23
Integrate OpenAI

Dev-Khant:
Resolves https://github.com/langfuse/langfuse/issues/187


Dev-Khant:
Here I think it would be better if we could add all the other libraries and frameworks in `integrations.py`

langfuse/integrations.py
```diff
@@ -0,0 +1,83 @@
...
+            (ChatCompletion, "create"),
+            (Completion, "create"),
+        ]
+
+        for api_resource_class, method in api_resources_classes:
+            create_method = getattr(api_resource_class, method)
+            setattr(api_resource_class, method, self.langfuse_modified(create_method))
+
+
+modifier = OpenAILangfuse()
```
maxdeichmann, line 82: why does this need to be executed here?

tests/test_integrations.py
```diff
@@ -0,0 +1,16 @@
...
+def test_openai_chat_completion():
+    completion = openai.ChatCompletion.create(
+        model="gpt-3.5-turbo", messages=[{"role": "user", "content": "1 + 1 = "}], temperature=0
+    )
+    assert len(completion.choices) != 0
+
+
+def test_openai_completion():
+    completion = openai.Completion.create(model="gpt-3.5-turbo-instruct", prompt="1 + 1 = ", temperature=0)
+    assert len(completion.choices) != 0
```
maxdeichmann, line 16: can you expose the flush function and call it in the tests? Also, did you try running this locally?

maxdeichmann:
```
    api = FintoLangfuse(
        username=os.environ.get("LF_PK"), password=os.environ.get("LF_SK"), environment=os.environ.get("HOST")
    )
    api.generations.get()
```
could you also fetch all the generations and check that the one you generate is there? Maybe adding a UUID as a prompt helps to find it?

langfuse/integrations.py
```diff
@@ -0,0 +1,83 @@
...
+
+import openai
+from openai.api_resources import ChatCompletion, Completion
+
+from langfuse import Langfuse
+from langfuse.client import InitialGeneration
+from langfuse.api.resources.commons.types.llm_usage import LlmUsage
+
+
+load_dotenv()
```
maxdeichmann, line 14: Why is this needed here? So far this worked without

langfuse/integrations.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+class OpenAILangfuse:
```
maxdeichmann, line 17: I would maybe rename the class to reflect a bit better of how it works.

langfuse/integrations.py
```diff
@@ -0,0 +1,83 @@
...
+                call_details = self._get_call_details(result, **kwargs)
+                call_details["startTime"] = startTime
+            except Exception as ex:
+                raise ex
+
+            return self._log_result(result, call_details)
+
+        return wrapper
+
+    def set_openai_funcs(self):
```
maxdeichmann, line 71: maybe `replace_openai_funcs`?

tests/test_integrations.py
Dev-Khant: I tested it locally and on UI it was showing properly @maxdeichmann

langfuse/integrations.py
Dev-Khant: As I have a PR #100 so it will be useful there. And in `.env` we will also have OpenAI_key so load that as well.

langfuse/integrations.py
Dev-Khant: Okay

langfuse/integrations.py
Dev-Khant: Do you suggest any?

tests/test_integrations.py
Dev-Khant: > can you expose the flush function and call it in the tests? Also, did you try running this locally?

I'm actually doing it inside a method. Do I need to call it again in tests?

```  
def _log_result(self, result, call_details):
      generation = InitialGeneration(name="OpenAI-generation", **call_details)
      self.langfuse.generation(generation)
      self.langfuse.flush()
      return result
```

Dev-Khant: > ```
>     api = FintoLangfuse(
>         username=os.environ.get("LF_PK"), password=os.environ.get("LF_SK"), environment=os.environ.get("HOST")
>     )
>     api.generations.get()
> ```
> 
> could you also fetch all the generations and check that the one you generate is there? Maybe adding a UUID as a prompt helps to find it?

It's throwing this `Pydantic` error. But I have made changes to include custom `trace_name` while making the call. Can check the last commit @maxdeichmann 
image

maxdeichmann: > I'm actually doing it inside a method. Do I need to call it again in tests?

ahh awesome, did not see that!

maxdeichmann: I fixed it here: https://github.com/langfuse/langfuse/commit/528f9ee2bf01640408394df14aa37e3b79686440

langfuse/integrations.py
maxdeichmann: Sure: `LangfuseOpenAiWrapper`

maxdeichmann: WDYT?

langfuse/integrations.py
maxdeichmann: I left one comment there. Happy to merge that asap

langfuse/integrations.py
Dev-Khant: Sounds good

langfuse/integrations.py
Dev-Khant: Great made changes there

tests/test_integrations.py
```diff
@@ -0,0 +1,40 @@
...
+api = FintoLangfuse(
+    environment=os.environ["HOST"],
+    username=os.environ["LF_PK"],
+    password=os.environ["LF_SK"],
+)
+
+
+def test_openai_chat_completion():
+    trace_id = create_uuid()
+    completion = openai.ChatCompletion.create(
```
maxdeichmann, line 21: I just tested this out locally. openai is shown on my end as type any in the IDE, whereas when using the original openai SDK, i get autocompletions. Is there a way to solve that?

langfuse/integrations.py
```diff
@@ -0,0 +1,99 @@
...
+            (ChatCompletion, "create"),
+            (Completion, "create"),
+        ]
+
+        for api_resource_class, method in api_resources_classes:
+            create_method = getattr(api_resource_class, method)
+            setattr(api_resource_class, method, self.langfuse_modified(create_method))
+
+
+modifier = OpenAILangfuse()
```
maxdeichmann, line 98: Do you think we can instantiate this as a Singleton? I fear our users will import this multiple times and then create multiple Langfuse objects under the hood. 
I think one instantiation of the Langfuse object would be ideal. Could we also expose the flush function from there? If yes, we can remove it above in the `_log_result` method and thereby save latencies for our users

maxdeichmann: Other than this, i think we are good to go. I just tested it on my machine also - seems to work well :)

tests/test_integrations.py
Dev-Khant: I didn't understand exactly, what's the issue @maxdeichmann

langfuse/integrations.py
maxdeichmann: image

`create` is of type any here. If you import `import openai` into `integrations.py`, this should be typed. Could you add that?

tests/test_integrations.py
maxdeichmann: image

create is of type `any` here. If you import `import openai` into `integrations.py`, this should be typed. Could you add that?

langfuse/integrations.py
maxdeichmann: Thanks a lot for the change on the singleton! What do you think of concurrency? I just read this here: https://stackabuse.com/creating-a-singleton-in-python/. 
Also, can you add a flush functionality to the module? Otherwise we hide the langfuse object and our users are not able to flush before the program exits.

langfuse/integrations.py
```diff
@@ -0,0 +1,99 @@
+import os
+import functools
+from datetime import datetime
+from dotenv import load_dotenv
+
+import openai
+from openai.api_resources import ChatCompletion, Completion
```
fxmb, line 7: @maxdeichmann we should probably not fully import "lesser known" packages directly to provide more context. Can we change this to `from openai import api_resources` and then call like `api_resources.ChatCompletion`? Just at thought

langfuse/integrations.py
```diff
@@ -0,0 +1,99 @@
...
+    def get_langfuse_args(self):
+        return {**self.args, **self.kwargs}
+
+    def get_openai_args(self):
+        return self.kwargs
+
+
+class OpenAILangfuse:
+    def __init__(self):
+        self.langfuse = Langfuse(os.environ["LF_PK"], os.environ["LF_SK"], os.environ["HOST"])
```
fxmb: This will throw an error if key is not present afaik. Is this intended? Rather use `getenv()` and provide fallback logic in case key is not present? It's a personal preference so feel free to ignore.

langfuse/integrations.py
Dev-Khant: This could help here https://medium.com/analytics-vidhya/how-to-create-a-thread-safe-singleton-class-in-python-822e1170a7f6. I have made the change

langfuse/integrations.py
Dev-Khant: Hi, If keys are not present it is handled in `Langfuse` class 
image

tests/test_integrations.py
Dev-Khant: Here @maxdeichmann for me its same, so how do I fix it
image
image

langfuse/integrations.py
Dev-Khant: I don't find a way to directly calling `flush` using openai. In order to use it we need to create a object of `OpenAILangfuse`. Do you need to do that??

maxdeichmann:
```
    langfuse = Langfuse()
    completion = openai.ChatCompletion.create(
        langfuse=langfuse, model="gpt-3.5-turbo", messages=[{"role": "user", "content": "1 + 1 = "}], temperature=0
    )
    langfuse.flush()
```

What do you think about such an experience? If we do that, we can also allow more complex scenarios in the future:
```
    langfuse = Langfuse()

    span = langfuse.span(...)

    completion = openai.ChatCompletion.create(
        langfuse=span, model="gpt-3.5-turbo", messages=[{"role": "user", "content": "1 + 1 = "}], temperature=0
    )
    langfuse.flush()
```

What do you think about that?

maxdeichmann:
```
    langfuse = Langfuse()
    completion = openai.ChatCompletion.create(
        langfuse=langfuse, model="gpt-3.5-turbo", messages=[{"role": "user", "content": "1 + 1 = "}], temperature=0
    )
    langfuse.flush()
```


What do you think about such an experience? If we do that, we can also allow more complex scenarios in the future:
```
    langfuse = Langfuse()

    span = langfuse.span(...)

    completion = openai.ChatCompletion.create(
        langfuse=span, model="gpt-3.5-turbo", messages=[{"role": "user", "content": "1 + 1 = "}], temperature=0
    )
    langfuse.flush()
```
What do you think about that?

Dev-Khant:
@maxdeichmann Tests are failing because Openai key is not found

langfuse/integrations.py
maxdeichmann: @Dev-Khant sorry if the above was not shown to you. What do you think of the above?

Dev-Khant: Sorry @maxdeichmann didn't saw the above comment. It's a useful approach but the `languse` inside openai wrapper will flush the inner processes and while the outer `langfuse` will handle processes defined out the class. For eg. processes like as you showed here `span = langfuse.span(...)`. So there won't be any conflict with `flush()` being used by two different instances of languse, right?

Dev-Khant:
> @maxdeichmann Tests are failing because Openai key is not found

And @maxdeichmann Please do have a look for this as well. Thanks

langfuse/integrations.py
maxdeichmann: No worries :)

Yes, you can create two different `Langfuse` objects in theory, they wont interact with each other. However, each of these will create their own worker thread to take care of network requests. I dont think this would be great for the performance of our users applications.
Also, the above approach gives us way more flexibility using the integration. Users could create generations in sub-spans of their trace which would not be possible without.

Dev-Khant: @maxdeichmann Sure then we could go with this approach!

maxdeichmann:
> > @maxdeichmann Tests are failing because Openai key is not found
> 
> And @maxdeichmann Please do have a look for this as well. Thanks

Yes, currently running all tests with missing API keys only locally. You can check the same behavior in the langchain tests and you can do the same.

Dev-Khant:
@maxdeichmann I have made the changes can you please run the tests again? Thanks

langfuse/integrations.py
maxdeichmann: I think before merging we should have that in here. I can take that over on the weekend.

Dev-Khant: Sure, let me know if you need my help.

PR #112, opened 2023-09-28
fix python versions

maxdeichmann:


derino:
Thanks for the quick fix!

PR #114, opened 2023-10-02
Inject additional headers

yassineselmi:
# Context

The goal of this change is to enable injecting custom headers to be sent as part of the API request.  Ot is useful when Langfuse is deployed behind a proxy that validates some access policies (for ex: Cloudflare Tunnel). Many other use cases can be covered by this change as well.

# Design

Optionally, the user can specify additional headers as follow:

```python
from langfuse.client import Langfuse

langfuse = Langfuse(
        public_key=env.LANGFUSE_PUBLIC_KEY,
        secret_key=env.LANGFUSE_SECRET_KEY,
        host=env.LANGFUSE_HOST,
       
        # Adding custom headers
        additional_headers = {
            # For example: Inject Cloudflare Service Token
            'CF-Access-Client-Id': env.LANGFUSE_CF_ACCESS_CLIENT_ID,
            'CF-Access-Client-Secret': env.LANGFUSE_CF_ACCESS_CLIENT_SECRET,
        }
)

```

### DIFF
--- a/langfuse/api/client.py
+++ b/langfuse/api/client.py
@@ -22,7 +22,8 @@ class FintoLangfuse:
         x_langfuse_sdk_version: typing.Optional[str] = None,
         x_langfuse_public_key: typing.Optional[str] = None,
         username: str,
-        password: str
+        password: str,
+        additional_headers: typing.Optional[typing.Dict] = {}
     ):
         self._environment = environment
         self.x_langfuse_sdk_name = x_langfuse_sdk_name
@@ -30,6 +31,7 @@ class FintoLangfuse:
         self.x_langfuse_public_key = x_langfuse_public_key
         self._username = username
         self._password = password
+        self.additional_headers = additional_headers
         self.dataset_items = DatasetItemsClient(
             environment=self._environment,
             x_langfuse_sdk_name=self.x_langfuse_sdk_name,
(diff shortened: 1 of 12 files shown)


maxdeichmann:
Hi @yassineselmi , thanks so much for the contribution! 

The situation is a bit tricky: We use Fern for the generation of our Python code. It has been super helpful so far generating boilerplate code for us. Much of this code was touched by your PR. Unfortunately, they do not allow to have a list of headers in their configurations: https://buildwithfern.com/docs/definition/api-yml#api-wide-headers. 
I fully understand your request though. What about we add one header to the SDK that you can use for your use case? Unfortunately this would already have a pre-defined naming. Would that work for you?

yassineselmi:
The static-name headers solve my issue for sure, but I'm wondering if this will be considered as a useful feature for others :p .

So, you mean that you are going to add:
`CF-Access-Client-Id` and `CF-Access-Client-Secret` headers?

maxdeichmann:
> The static-name headers solve my issue for sure, but I'm wondering if this will be considered as a useful feature for others :p .
> 
> So, you mean that you are going to add: `CF-Access-Client-Id` and `CF-Access-Client-Secret` headers?

Yes, you are right. We are currently considering how to make the Python SDK more usabel (Pydantic models, Headers, and more. Would you also be fine with different names? We could for example add something like `X-Langfuse-User-Defined-1`, `X-Langfuse-User-Defined-2`, `X-Langfuse-User-Defined-1` as an intermediary solution until we revamped this SDK. Would different names than the ones described by you work?

marcklingen:
unfortunately all parts within /api cannot be modified as they are generated by https://github.com/fern-api/fern

happy to add this once fern supports custom header passthrough

PR #119, opened 2023-10-09
Llamaindex integration

Dev-Khant:


lucasiscovici:
Hello any updates about this integration ? :)

maxdeichmann:
> Hello any updates about this integration ? :)

@lucasiscovici thanks for raising this. We have it in our backlog but first need to solve some other issues we have. We will let you know when this is done.

hubert-springbok:
We're interested in this callback as well. In fact, we missed this one and created our own version -_-

One note: for these kind of third-party integrations, you'll want to avoid having the third-party as an explicit dependency, at least by default. Otherwise by the time you have 30 callbacks, langfuse would depend on half of PyPi.

Here is an example from [sentry-sdk](https://github.com/getsentry/sentry-python/blob/master/setup.py#L46), which specifies third-party ranges it can work with. The end user says `pip install sentry-sdk[fastapi]` etc., according to what they actually need. In the corresponding integration code, they [use try-catch](https://github.com/getsentry/sentry-python/blob/0eb346533da224f2d6d99c87e06be5e26eaa5cf1/sentry_sdk/integrations/fastapi.py#L23)

Dev-Khant:
> We're interested in this callback as well. In fact, we missed this one and created our own version -_-
> 
> One note: for these kind of third-party integrations, you'll want to avoid having the third-party as an explicit dependency, at least by default. Otherwise by the time you have 30 callbacks, langfuse would depend on half of PyPi.
> 
> Here is an example from [sentry-sdk](https://github.com/getsentry/sentry-python/blob/master/setup.py#L46), which specifies third-party ranges it can work with. The end user says `pip install sentry-sdk[fastapi]` etc., according to what they actually need. In the corresponding integration code, they [use try-catch](https://github.com/getsentry/sentry-python/blob/0eb346533da224f2d6d99c87e06be5e26eaa5cf1/sentry_sdk/integrations/fastapi.py#L23)

Thanks @hubert-springbok for pointing out. We will look into it.

Dev-Khant:
@maxdeichmann @marcklingen Should I continue with this?

hubert-springbok:
Two cents:

1. We have some further work on our end around having a single trace that contains both 'manual' LLM calls and LlamaIndex pipelines. Happy to discuss.

2. Note that the extra approach described above has been since rolled out for the langchain callback, so it should be easy to follow now, see #214

maxdeichmann:
Hi @Dev-Khant thanks a lot for the contribution here! We worked on a llamaindex integration ourselves and looked into your changes for inspiration. We will launch this today more broadly.

Dev-Khant:
Thank you @maxdeichmann for giving me a chance at this.

PR #164, opened 2023-11-13
Set userId in langfuse trace if "user" is passed in OpenAI completion API calls

code-haven:
Use the "user" parameter in https://platform.openai.com/docs/guides/safety-best-practices/end-user-ids to be able to group usage by user in https://langfuse.com/docs/user-explorer.


maxdeichmann:
@code-haven thanks for the contribution. I am happy to push this out asap. Do you mind renaming `user` to `user_id`? Some of our users might insert an object to `user` which would not work. With `user_id` we have higher likelihood of getting strings :)

code-haven:
@maxdeichmann done

code-haven:
oops, looks like the OpenAI library isn't liking the user_id parameter. Have to test if this work again.

maxdeichmann:
> oops, looks like the OpenAI library isn't liking the user_id parameter. Have to test if this work again.

Ok. Is it ok to have it the way it is now? Having the userId on the generation function might be confusing for users.

code-haven:
are you suggesting to create a custom trace with userId and passing trace_id to the chatCompletion call instead of making this change?

maxdeichmann:
> eate a custom trace with userId and passing trace_id to the chatCompletion call instead of making this change (if we leave this as it is now)?

Yes, i think this would be the cleanest solution for now.

code-haven:
ok, sounds good. thanks for the help!

PR #206, opened 2023-11-30
feat(core): on .create, .span, .event, add start times automatically and allow to add body to .end methods.

maxdeichmann:


nfsergiu:
@maxdeichmann Hey! I was wondering why are you setting the current timestamp if `start_time` is not set by the user? As far as I understand langfuse will accept a null value for `startTime` and consequently use the timestamp at insert time.

This does not allow a user of the SDK to do a **real** upsert, since every upsert (e.g. `langfuse.generation(id="previously_set_id", output="some random output")` since it will automatically set a `start_time` which will end up in langfuse updating the old `start_time` which was set when the generation was first created.

Currently there is no way to achieve the behaviour I want.
I think users should have the option to NOT pass a start_time parameter and keep the same timestamp from when the generation was created. Maybe have the option of passing a Literal like `unchanged` or something similar.

PR #208, opened 2023-11-30
feat(core): Remove pydantic from SDK interfaces

maxdeichmann:


langfuse/callback.py
```diff
@@ -198,51 +197,41 @@ def __generate_trace_and_parent(
...
+                        metadata=self.__join_tags_and_metadata(tags, metadata),
+                        input=inputs,
+                        start_time=datetime.now(),
+                        version=self.version,
                     )
                     if self.rootSpan is None
                     else self.rootSpan.span(
-                        CreateSpan(
-                            id=self.nextSpanId,
-                            traceId=self.trace.id,
```
maxdeichmann, line 239: this was unnecessary. TraceId will be taken care off by the SDK

langfuse/callback.py
```diff
@@ -467,15 +452,15 @@ def on_tool_error(
...
+            self.runs[run_id] = self.runs[run_id].update(status_message=error, level=ObservationLevel.ERROR, end_time=datetime.now(), version=self.version)
         except Exception as e:
             self.log.exception(e)
 
     def __on_llm_action(
         self,
         serialized: Dict[str, Any],
         run_id: UUID,
-        prompts: List[str],
+        prompts: Union[List[str], List[List[BaseMessage]]],
```
maxdeichmann, line 463: this was wrong originally

langfuse/task_manager.py
```diff
@@ -146,16 +146,19 @@ def init_resources(self):
             consumer.start()
             self._consumers.append(consumer)
 
-    def add_task(self, event):
+    def add_task(self, event: dict):
         try:
-            self._log.debug("Adding task")
+            self._log.debug(f"adding task {event}")
+            json.dumps(event, cls=EventSerializer)
```
maxdeichmann, line 152: if event is not serializable, it will log an error to the console without causing an exception.

PR #216, opened 2023-12-07
fix: Langfuse and StatefulClient clients return incorrect trace url

HugoMichard:
Added project id in trace url returned by Langfuse client and StatefulClient

Closes https://github.com/langfuse/langfuse/issues/650

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,7 @@ jobs:
   ci:
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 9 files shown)


marcklingen:
Hi @HugoMichard, thanks for the contribution. Does this fix a problem that you have?

The /traces/{traceId} route is redirected to the correct /project/{projectId}/traces/{traceId} route to not require a synchronous call to the Langfuse API to get the trace url. See implementation here: https://github.com/langfuse/langfuse/blob/main/src/pages/trace/%5BtraceId%5D.tsx

PR #219, opened 2023-12-07
Fix for mapping error in LlmUsage with ChatOllama and ChatLiteLLM clients

DavidLMS:
This Pull Request addresses a critical issue identified following the implementation of PR #150 in the langfuse-python library. The problem occurs when using clients such as ChatOllama or ChatLiteLLM, leading to a `TypeError` when trying to unpack a `Usage` object into the `LlmUsage` class.

**Identified Error:**
```
langfuse.api.resources.commons.types.llm_usage.LlmUsage() argument after ** must be a mapping, not Usage
Traceback (most recent call last):
File "/opt/homebrew/lib/python3.11/site-packages/langfuse/callback.py", line 596, in on_llm_end
llm_usage = None if response.llm_output is None else LlmUsage(**response.llm_output["token_usage"])
```

**Proposed Solution:**
I have modified the code to ensure that the `token_usage` object is an instance of `Usage` before being assigned to `llm_usage`. This change ensures the correct assignment of the `token_usage` object to `llm_usage`, thus avoiding the mapping error that was previously occurring.


DavidLMS:
Resolved in other PR.

PR #223, opened 2023-12-10
feat(api): add support for the new usage object

maxdeichmann:


maxdeichmann:
Tests failing as we are testing against the wrong docker image

PR #227, opened 2023-12-11
Remove unnecessary dependencies

hubert-springbok:
A follow-up on the quest of making this a lightweight SDK.

Removed the following core dependencies:
 * `monotonic`: only useful with Python < 3.3. This project targets 3.8+
 * `attrs`: there is no usage at all across the repository. Only transitive dependency is `langchain`.
 * `python_dateutil`: used in one place only, where builtin functionality is sufficient
 * `pytz`: only used in one test; builtin functionality is sufficient

Code to check the python_dateutil replacement is sound:
```python
before = datetime.utcnow().replace(tzinfo=tzutc())
after = datetime.now(timezone.utc)
difference = before - after
print(difference)
```


maxdeichmann:
@hubert-springbok did you run the tests locally?

hubert-springbok:
Hi, I ran the `tests/test_core_sdk.py` and all of them passed.

I've also run the full collection, but two of them fail because they expect
some additional setup I couldn't get right on localhost. I believe these
also fail on `main` for me.

That's the pytest summary:
```
========================================================================
short test summary info
========================================================================
FAILED tests/test_langchain.py::test_callback_openai_functions_python -
AssertionError: assert 'https://cloud.langfuse.com' == '
http://localhost:3000'
FAILED tests/test_langchain.py::test_create_extraction_chain -
langfuse.api.resources.commons.errors.not_found_error.NotFoundError:
status_code: 404, body: {'message': 'Trace not found within authorized
project'}
================================================== 2 failed, 112 passed, 6
skipped, 881 warnings in 193.27s (0:03:13)
===================================================```

Is there a problem with the tests?

> Message ID: ***@***.***>
>

**Comment by hubert-springbok** [comment]
I believe the errors in CI/CD are due to missing environment variable. Locally I set the public and private langfuse keys.
The CI/CD complains:
```
FAILED tests/test_sdk_setup.py::test_wrong_url_auth_check - ValueError: public_key is required, set as parameter or environment variable 'LANGFUSE_PUBLIC_KEY'
```

You may need to change something in GitHub configuration to allow this MR to access secrets, but I don't have enough experience to know what to look for.

hubert-springbok:
Hi @maxdeichmann , I've merged main back into this and resolved conflicts.
It seems the CI/CD tests failed in the same way on #214 as well, which you've managed to merge before.

Any idea on how we can resolve this?

hubert-springbok:
I'm happy to submit a MR where the changes are very close to zero, just to rule out that this MR is actually to blame for the CI/CD failure, if that helps

maxdeichmann:
Hi @hubert-springbok, just looking into this. Sorry for the delay. In https://github.com/langfuse/langfuse-python/pull/214, we ran into missing dependencies for some users. Hence not sure what exactly causes the issues here.

The last time i had to provide this fix here, as langchain was used in the main app while it was optional: https://github.com/langfuse/langfuse-python/commit/f538b39be843b81928938a8791cc2140721175ba

hubert-springbok:
> Hi @hubert-springbok, just looking into this. Sorry for the delay. In #214, we ran into missing dependencies for some users. Hence not sure what exactly causes the issues here.
> 
> The last time i had to provide this fix here, as langchain was used in the main app while it was optional: [f538b39](https://github.com/langfuse/langfuse-python/commit/f538b39be843b81928938a8791cc2140721175ba)

Yes, unfortunately that's the problem with plugins and opitonal dependencies. Ideally there is few places in your core code that has special behaviour depending on third-party objects and you can bypass them manually like you did.

In this MR I aim to completely remove the need to depend on the packages I've uninstalled, so users should not experience problems like with langchain plugin.

maxdeichmann:
> > Hi @hubert-springbok, just looking into this. Sorry for the delay. In #214, we ran into missing dependencies for some users. Hence not sure what exactly causes the issues here.
> > The last time i had to provide this fix here, as langchain was used in the main app while it was optional: [f538b39](https://github.com/langfuse/langfuse-python/commit/f538b39be843b81928938a8791cc2140721175ba)
> 
> Yes, unfortunately that's the problem with plugins and opitonal dependencies. Ideally there is few places in your core code that has special behaviour depending on third-party objects and you can bypass them manually like you did.
> 
> In this MR I aim to completely remove the need to depend on the packages I've uninstalled, so users should not experience problems like with langchain plugin.

Yes, i think my last commit solved an issue. Will deploy this in an alpha release and we can see whether it works.

PR #299, opened 2024-01-22
Fix Issue #598

shashankshet:
Added exception handling when env variables are not present, logging the error instead of throwing exception


shashankshet:
Hi @maxdeichmann ,

 the unit test cases when run individually are passing the test cases, but while I use the pytest command
` poetry run pytest -s -v --log-cli-level=INFO tests/test_sdk_setup.py`
few of the test cases fail.
Can you correct me if the envi variable values which I am using are correct?

 LANGFUSE_HOST=https://cloud.langfuse.com
 LANGFUSE_PUBLIC_KEY=pk-lf-1234567890
 LANGFUSE_SECRET_KEY=sk-lf-1234567890

maxdeichmann:
> Hi @maxdeichmann ,
> 
> 
> 
>  the unit test cases when run individually are passing the test cases, but while I use the pytest command
> 
> ` poetry run pytest -s -v --log-cli-level=INFO tests/test_sdk_setup.py`
> 
> few of the test cases fail.
> 
> Can you correct me if the envi variable values which I am using are correct?
> 
> 
> 
>  LANGFUSE_HOST=https://cloud.langfuse.com
> 
>  LANGFUSE_PUBLIC_KEY=pk-lf-1234567890
> 
>  LANGFUSE_SECRET_KEY=sk-lf-1234567890
> 
> 

This looks good but will execute the tests against our hosted server.
Alternatively, you can run the main project on your machine and use localhost:3000 as host. Username/password remain the same. 

If all tests after the test file are failing, you might remove an env variable in the test and don't add it back, so that all remaining tests fail.

maxdeichmann:
Hi @shashankshet thanks a lot for the contribution!
When API keys are missing and we are not throwing exceptions, all downstream functionality of the SDK will fail and users will see lots of error messages. If the API keys are non existent, we should basically turn off all tracing capabilities of the SDK. This is a bigger effort and will be prioritized by us.

PR #326, opened 2024-02-02
feat: adds prompt caching

hassiebp:
## Problem

Currently, on every call to `langfuse.get_prompt(*)` a GET request is initiated to the `/prompts` endpoint. This PR implements SDK-level caching of prompts. Cache TTL can be set via `cache_ttl_seconds` property in an optional third options param in `langfuse.get_prompt`. If not specified, a default cache TTL is used (currently 60 seconds).

In case there is an expired prompt cache but fetching the fresh prompt fails, the stale prompt will be returned from `langfuse.get_prompt` to enable failover.

## Changes

- Introduces prompt caching to the Python SDK
- Default cache TTL set to 60 seconds
- Enables failover to stale prompt cache when fetching the fresh prompt fails
- Optional third param added to `langfuse.get_prompt(*)` to customize cache TTL through the `cache_ttl_seconds` property

### Bump level

- [ ] Major
- [X] Minor
- [ ] Patch

### Changelog notes

- Added support for SDK-level caching of prompts


maxdeichmann (commented):
Thanks for the tests. They give lots of confidence here.

maxdeichmann (approved):

PR #332, opened 2024-02-05
fix: Fix httpx.client not initialized if LangfuseClient does not have…

brandonkzw:
Fix failing pytest when LangfuseClient not initialized with session

### DIFF
--- a/langfuse/request.py
+++ b/langfuse/request.py
@@ -23,7 +23,7 @@ class LangfuseClient:
         base_url: str,
         version: str,
         timeout: int,
-        session: httpx.Client = httpx.Client(),
+        session: httpx.Client,
     ):
         self._public_key = public_key
         self._secret_key = secret_key
--- a/langfuse/request.py
+++ b/langfuse/request.py
@@ -23,7 +23,7 @@ class LangfuseClient:
         base_url: str,
         version: str,
         timeout: int,
-        session: httpx.Client,
+        session: httpx.Client = httpx.Client(),
     ):
         self._public_key = public_key
         self._secret_key = secret_key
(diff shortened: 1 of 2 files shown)


langfuse/request.py
```diff
@@ -23,7 +23,7 @@ def __init__(
         base_url: str,
         version: str,
         timeout: int,
-        session: httpx.Client,
+        session: httpx.Client = httpx.Client(),
```
maxdeichmann, line 26: why is this change required?

brandonkzw: the pytest fix commit you have done https://github.com/langfuse/langfuse-python/commit/9fde0df725d077ee4929ea53f260c47a6946f630 shows that if LangfuseClient are initialized without session will cause an issue, previously it was relying on `_session = httpx.Client()` https://github.com/langfuse/langfuse-python/blob/c06b2927733f99dff5e98c3fc1599a3d627255f3/langfuse/request.py#L10

maxdeichmann: Yes, i think that outside of tests we are fine with your change here:

https://github.com/langfuse/langfuse-python/commit/389da0c5be4e916fea9e896331682e5d03520dc3#diff-e3c134307c29183eb3f5860e662771700bb0486cb360ea9facfad0081cc56d44R104

This is the only place in the code where we initialise FernLangfuse and LangfuseClient except for test code.

maxdeichmann: If you agree, feel free to remove this line and we should merge this PR.

brandonkzw: Agreed, removed line :)

maxdeichmann: Thanks a lot!

PR #357, opened 2024-02-13
fix: support more types for JSON serialization

SinghCoder:
Fixes https://github.com/langfuse/langfuse/issues/948

For my case, it was a dataclass which I was returning as a chain output, and python default JSON serializer does not support dataclasses dumping.. See [StackOverflow Thread](https://stackoverflow.com/questions/51286748/make-the-python-json-encoder-support-pythons-new-dataclasses).

So, I have fixed that as per answer in the thread.

But, there are many more cases where it might fail later also.. So, I checked how LangSmith folks have done.. Here is the [source](https://github.com/langchain-ai/langsmith-sdk/blob/6ff1369f7c226e79782fbe51106d67fbe466218f/python/langsmith/client.py#L158).. Maybe we can take some more inspirations from there, and make the serializer more robust here...

### DIFF
--- a/langfuse/serializer.py
+++ b/langfuse/serializer.py
@@ -1,6 +1,8 @@
 from datetime import date, datetime
+from dataclasses import is_dataclass, asdict
 from json import JSONEncoder
 from typing import Any
+from uuid import UUID
 
 from langfuse.api.core import serialize_datetime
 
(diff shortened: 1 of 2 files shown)


maxdeichmann:
@SinghCoder thanks a lot for the contribution. I will test this and get this live as soon as possible.

PR #367, opened 2024-02-14
Add SDK integration examples for fastAPI and django

ChrisTho23:


examples/fastapi_example/fastapi_example/main.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@app.get("/")
+async def main_route():
+    return {"message": "Hey, this is an example showing how to use Langfuse with FastAPI."}
+
+# Initialize Langfuse
+langfuse = Langfuse(public_key="pk-lf-1234567890", secret_key="sk-lf-1234567890", host="http://localhost:3000")
+
+async def get_response_openai(prompt, background_tasks: BackgroundTasks):
```
maxdeichmann, line 30: I think we can remove the function parameters here, right?

examples/fastapi_example/pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+version = "0.1.0"
+description = ""
+authors = ["ChrisTho23 <christophe.thomassin23@gmail.com>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.10"
+fastapi = "^0.109.2"
+uvicorn = "^0.27.1"
+langfuse = "2.13.3a0"
```
maxdeichmann, line 12: can you use latest langfuse here always?

examples/django_example/myapp/views.py
```diff
@@ -0,0 +1,11 @@
+from django.shortcuts import render
+from django.http import JsonResponse, HttpResponseServerError
+from myapp.langfuse_integration import get_response_openai, langfuse
```
maxdeichmann, line 3: i think here are some unused imports

maxdeichmann (commented):
I am wondering whether we can remove some of the empty django files to make the example more readable?

ChrisTho23:
> I am wondering whether we can remove some of the empty django files to make the example more readable?

Thought the same thing. They are autogenerated so was not sure if one traditionally keeps them. I can remove them and check if it still works.

examples/django_example/myapp/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+    This function handles the shutdown process.
+    
+    It calls the langfuse_flush function to flush any pending changes,
+    and then exits the program with a status code of 0.
+    """
+    langfuse_flush()
+    sys.exit(0)
+
+signal.signal(signal.SIGINT, shutdwon_handler)
```
maxdeichmann, line 16: I think we can also listen to sigterm here as this is used to gracefully terminate a process. https://www.baeldung.com/linux/sigint-and-other-termination-signals

ChrisTho23:
@maxdeichmann implemented your comments. Did not remove examples/django_example/myapp/migrations bc it was actually used for some migrations and thus, could also be helpful for users.

PR #384, opened 2024-02-19
feat: adds llama-index integration

hassiebp:
Adds llama-index integration to Python SDK.

- Adds LLamaIndexCallbackHandler
- Refactors shared code between `LangchainCallbackHandler` and `LlamaIndexCallbackHandler` to `BaseCallbackHandler`
- Adds class and method decorators to function as error boundaries for errors to be logged and not reraised at the highest level

### DIFF
--- a/langfuse/callback/llama_index.py
+++ b/langfuse/callback/llama_index.py
@@ -88,12 +88,12 @@ class LLamaIndexCallbackHandler(
 
         self.root = stateful_client
         self.event_map: Dict[str, List[CallbackEvent]] = defaultdict(list)
-        self._cur_trace_id: Optional[str] = None
+        self._llama_index_trace_name: Optional[str] = None
         self._token_counter = TokenCounter(tokenizer)
 
     def start_trace(self, trace_id: Optional[str] = None) -> None:
         """Run when an overall trace is launched."""
-        self._cur_trace_id = trace_id
+        self._llama_index_trace_name = trace_id
 
     def end_trace(
         self,
@@ -105,6 +105,9 @@ class LLamaIndexCallbackHandler(
             self.log.debug("No events in trace map to create the observation tree.")
             return
 
+        # Generate Langfuse observations after trace has ended and full trace_map is available.
+        # For long-running traces this leads to events only being sent to Langfuse after the trace has ended.
+        # Timestamps remain accurate as they are set at the time of the event.
         self._create_observations_from_trace_map(
             event_id=BASE_TRACE_EVENT, trace_map=trace_map
         )
@@ -169,7 +172,7 @@ class LLamaIndexCallbackHandler(
         else:
             self.trace = self.langfuse.trace(
                 id=str(uuid4()),
-                name=self.trace_name or f"LlamaIndex_{self._cur_trace_id}",
+                name=self.trace_name or f"LlamaIndex_{self._llama_index_trace_name}",
                 version=self.version,
                 session_id=self.session_id,
                 user_id=self.user_id,
@@ -219,7 +222,7 @@ class LLamaIndexCallbackHandler(
             elif EventPayload.MESSAGES in end_event.payload:
                 input = end_event.payload.get(EventPayload.MESSAGES)
                 response = end_event.payload.get(EventPayload.RESPONSE, {})
-                output = response.message
+                output = response.message.copy()
                 if hasattr(output, "additional_kwargs"):
                     delattr(output, "additional_kwargs")
                 model = response.raw.get("model", None)


langfuse/callback/llama_index.py
```diff
@@ -0,0 +1,366 @@
...
+                response = end_event.payload.get(EventPayload.RESPONSE, {})
+                output = response.message
+                model = response.raw.get("model", None)
+                token_usage = dict(response.raw.get("usage", {}))
+                usage = None
+                if token_usage:
+                    usage = {
+                        "prompt_tokens": token_usage.get("prompt_tokens"),
+                        "completion_tokens": token_usage.get("completion_tokens"),
+                        "total_tokens": token_usage.get("total_tokens"),
```
maxdeichmann, line 231: Placeholder: our usage object

langfuse/callback/llama_index.py
```diff
@@ -0,0 +1,366 @@
...
+                    }
+
+        generation = parent.generation(
```
maxdeichmann, line 234: We only start creating generations when this function is called?

langfuse/callback/llama_index.py
```diff
@@ -0,0 +1,366 @@
...
+            flush_interval=flush_interval,
+            max_retries=max_retries,
+            timeout=timeout,
+            sdk_integration="llama-index",
+        )
+
+        self.root = stateful_client
+        self.event_map: Dict[str, List[CallbackEvent]] = defaultdict(list)
+        self._cur_trace_id: Optional[str] = None
+        self._token_counter = TokenCounter(tokenizer)
```
maxdeichmann, line 92: Is this something our users need to provide?

pyproject.toml
```diff
@@ -7,14 +7,15 @@ license = "MIT"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = ">=3.8.1,<4.0"
+python = ">=3.8.1,<3.12"
```
maxdeichmann, line 10: :D

langfuse/decorators/error_logging.py
```diff
@@ -0,0 +1,43 @@
...
+    return wrapper
+
+
+def auto_decorate_methods_with(decorator):
+    """
+    Class decorator to automatically apply a given decorator to all
+    methods of a class.
+    """
+
+    def class_decorator(cls):
```
maxdeichmann, line 26: add condition that init is excluded from this.

langfuse/callback/llama_index.py
```diff
@@ -0,0 +1,366 @@
...
+        generation = parent.generation(
+            id=event_id,
+            trace_id=trace_id,
+            version=self.version,
+            name=name,
+            start_time=start_event.time,
+            end_time=end_event.time,
+            usage=usage or None,
+            model=model,
+            input=input,
```
maxdeichmann, line 243: chat format

maxdeichmann: Check if additional_kwargs is filled on input or output, then send it, otherwise do not send it.

langfuse/callback/llama_index.py
maxdeichmann: Set this to the Langfuse usage model (input, output, total). Langfuse then sets the correct unit type in case it is aware of the model.

langfuse/callback/llama_index.py
```diff
@@ -0,0 +1,366 @@
...
+
+        generation = parent.generation(
+            id=event_id,
+            trace_id=trace_id,
+            name=name,
+            start_time=start_event.time,
+            end_time=end_event.time,
+            version=self.version,
+            model=model,
+            input=chunks,
```
maxdeichmann, line 293: Remove the input, as this will be too large for our API

langfuse/callback/llama_index.py
```diff
@@ -0,0 +1,366 @@
...
+            input, output = self.handle_chunking_payload(self.event_map[event_id])
+
+        span = parent.span(
+            id=event_id,
+            trace_id=trace_id,
+            start_time=start_event.time,
+            name=start_event.event_type.value,
+            version=self.version,
+            session_id=self.session_id,
+            user_id=self.user_id,
```
maxdeichmann, line 328: there is no user_id on span

pyproject.toml
```diff
@@ -7,14 +7,15 @@ license = "MIT"
...
 httpx = ">=0.15.4,<0.26.0"
 pydantic = ">=1.10.7, <3.0"
 backoff = "^2.2.1"
 openai = ">=0.27.8"
 wrapt = "1.14"
 langchain = { version = ">=0.0.309", optional = true }
 chevron = "^0.14.0"
+llama-index = {version = "^0.10.6", optional = true}
```
maxdeichmann, line 18: Do you know by any chance ntil when this implementation is stable? It would be great to also support versions which are a couple of months old.

langfuse/callback/llama_index.py
```diff
@@ -0,0 +1,369 @@
...
+            max_retries=max_retries,
+            timeout=timeout,
+            sdk_integration="llama-index",
+        )
+
+        self.root = stateful_client
+        self.event_map: Dict[str, List[CallbackEvent]] = defaultdict(list)
+        self._cur_trace_id: Optional[str] = None
+        self._token_counter = TokenCounter(tokenizer)
+
```
maxdeichmann, line 93: I think it would be great to have variable declaraitons for these at the top at class level to help our linters.

langfuse/callback/llama_index.py
```diff
@@ -0,0 +1,369 @@
...
+    def start_trace(self, trace_id: Optional[str] = None) -> None:
+        """Run when an overall trace is launched."""
+        self._cur_trace_id = trace_id
```
maxdeichmann, line 96: Could you add a quick note that the trace_id is

langfuse/callback/llama_index.py
```diff
@@ -0,0 +1,369 @@
...
+            if EventPayload.PROMPT in end_event.payload:
+                input = end_event.payload.get(EventPayload.PROMPT)
+                output = end_event.payload.get(EventPayload.COMPLETION)
+
+            elif EventPayload.MESSAGES in end_event.payload:
+                input = end_event.payload.get(EventPayload.MESSAGES)
+                response = end_event.payload.get(EventPayload.RESPONSE, {})
+                output = response.message
+                if hasattr(output, "additional_kwargs"):
+                    delattr(output, "additional_kwargs")
```
maxdeichmann, line 224: I think we would want to keep the

langfuse/callback/llama_index.py
```diff
@@ -0,0 +1,369 @@
...
+        self,
+        trace_id: Optional[str] = None,
+        trace_map: Optional[Dict[str, List[str]]] = None,
+    ) -> None:
+        """Run when an overall trace is exited."""
+        if not trace_map:
+            self.log.debug("No events in trace map to create the observation tree.")
+            return
+
+        self._create_observations_from_trace_map(
```
maxdeichmann, line 108: can you add a short comment on why we do this at the end and the performance implications for index + query?

PR #400, opened 2024-02-21
feat(openai): add support for tags

marcklingen:


langfuse/openai.py
```diff
@@ -187,17 +189,24 @@ def _get_langfuse_data_from_kwargs(
     if user_id is not None and not isinstance(user_id, str):
         raise TypeError("user_id must be a string")
 
+    tags = kwargs.get("tags", None)
+    if tags is not None and (
+        not isinstance(tags, list) or not all(isinstance(tag, str) for tag in tags)
```
maxdeichmann, line 194: I think we have to use `List` here. See import on top.

PR #402, opened 2024-02-21
Update readme and prevent backoff from throwing L1 exceptions

ChrisTho23:


ChrisTho23:
Test file was renamed but exemplary test comment in readme not

langfuse/logging.py
```diff
@@ -10,3 +10,4 @@ def clean_logger():
     backoff_logger = logging.getLogger("backoff")
     backoff_logger.setLevel(logging.WARNING)  # Set the desired log level
     backoff_logger.addHandler(console_handler)
+    backoff_logger.propagate = False  # Stop propagation to the root logger
```
maxdeichmann, line 13: https://docs.python.org/3/library/logging.html#logging.Logger.propagate

I just found this here. If our users use backoff and hence import it as well, i think we will turn off logs for them too with this implementation.

hassiebp:
Thanks, Christophe - closing this as stale :)

PR #405, opened 2024-02-22
Cleanup and explicit type hints

hubert-springbok:
Follow up on #396 
 * remove a leftover exception print
 * move the langchain import check to the langchain plugin definition
 * add explicit type hints on client public API methods
 * fix misc type errors, e.g. `any` -> `Any`
 * minor clarity refactors, e.g. `x or y` instead of `x if x is not None else y`


langfuse/callback/langchain.py
```diff
@@ -1,4 +1,12 @@
 import logging
+
+log = logging.getLogger("langfuse")
+
+try:  # Test that langchain is installed before proceeding
+    import langchain
+except ImportError as e:
+    log.exception(f"Could not import langchain. Some functionality may be missing. {e}")
```
hubert-springbok, line 8: May want to use `log.error` instead. The stack trace is ugly and not interesting. Langchain is just not installed and that's it.
May want to raise a clear exception here anyways, since the code will hit another import error a few lines down. The functionality is most definitely missing.

langfuse/client.py
```diff
@@ -480,24 +476,21 @@ def score(
         id: typing.Optional[str] = None,
         comment: typing.Optional[str] = None,
         observation_id: typing.Optional[str] = None,
-        kwargs=None,
```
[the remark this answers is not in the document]
maxdeichmann: Yes, this is a bug.  Coudl you quickly add it?

hubert-springbok:
I can try to make an edit in the evening, but it's pretty busy here at the
moment and I'm OOO tomorrow.
You should be able to push here as well if I can't manage.

Hubert Bereś

hubert-springbok:
Ok, I got a minute now, the change is easy, have a look

langfuse/callback/langchain.py
maxdeichmann: Thanks!

PR #425, opened 2024-02-28
feat: adds decorator implementation

hassiebp:
Adds decorator implementation to Python SDK.

### DIFF
--- a/langfuse/decorators/langfuse_decorator.py
+++ b/langfuse/decorators/langfuse_decorator.py
@@ -250,19 +250,27 @@ class LangfuseDecorator:
                 {"args": logged_args, "kwargs": func_kwargs} if capture_input else None
             )
 
+            params = {
+                "id": id,
+                "name": name,
+                "start_time": start_time,
+                "input": input,
+            }
+
             # Create observation
             if parent and as_type == "generation":
-                observation = parent.generation(
-                    id=id, name=name, start_time=start_time, input=input
+                observation = parent.generation(**params)
+            elif as_type == "generation":
+                # Create wrapper trace if generation is top-level
+                # Do not add wrapper trace to stack, as it does not have a corresponding end that will pop it off again
+                trace = langfuse.trace(id=id, name=name, start_time=start_time)
+                observation = langfuse.generation(
+                    name=name, start_time=start_time, input=input, trace_id=trace.id
                 )
             elif parent:
-                observation = parent.span(
-                    id=id, name=name, start_time=start_time, input=input
-                )
+                observation = parent.span(**params)
             else:
-                observation = langfuse.trace(
-                    id=id, name=name, start_time=start_time, input=input
-                )
+                observation = langfuse.trace(**params)
 
             _observation_stack_context.set(stack + [observation])
 
--- a/tests/test_decorators.py
+++ b/tests/test_decorators.py
@@ -900,3 +900,29 @@ async def test_async_nested_openai_chat_stream():
     assert generation.usage.output is not None
     assert generation.usage.total is not None
     assert "2" in generation.output
+
+
+def test_generation_at_highest_level():
+    mock_trace_id = create_uuid()
+    mock_result = "Hello, World!"
+
+    @observe(as_type="generation")
+    def main():
+        return mock_result
+
+    result = main(langfuse_observation_id=mock_trace_id)
+    langfuse_context.flush()
+
+    assert result == mock_result
+
+    trace_data = get_api().trace.get(mock_trace_id)
+    assert (
+        trace_data.output is None
+    )  # output will be attributed to generation observation
+
+    # Check that the generation is wrapped inside a trace
+    assert len(trace_data.observations) == 1
+
+    generation = trace_data.observations[0]
+    assert generation.type == "GENERATION"
+    assert generation.output == result


langfuse/client.py
```diff
@@ -840,7 +841,7 @@ def generation(
         start_time: typing.Optional[dt.datetime] = None,
         end_time: typing.Optional[dt.datetime] = None,
         metadata: typing.Optional[typing.Any] = None,
-        level: typing.Optional[Literal["DEBUG", "DEFAULT", "WARNING", "ERROR"]] = None,
+        level: typing.Optional[SpanLevel] = None,
```
maxdeichmann, line 844: In general i like this change for DRY, but it has the downside, tha

langfuse/types/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+    name: Optional[str]
+    user_id: Optional[str]
+    session_id: Optional[str]
+    version: Optional[str]
+    release: Optional[str]
+    metadata: Optional[Any]
+    tags: Optional[List[str]]
+
+
+class ObservationParams(TraceMetadata, TypedDict):
```
maxdeichmann, line 17: Observations do not have session_id, tags, user_id

langfuse/decorators/langfuse.py
```diff
@@ -0,0 +1,388 @@
...
+            However, explicit calls to `flush` can be beneficial in certain edge cases or for debugging purposes.
+        """
+
+        langfuse = self._get_langfuse()
+        if langfuse:
+            langfuse.flush()
+        else:
+            self.log.warn("No langfuse object found in the current context")
+
+    def _get_langfuse(self) -> Langfuse:
```
maxdeichmann, line 367: This here might be called concurrently / parallel. I guess we shoudl make sure this behaves like a singleton. We have a similar implementation in the openai integration. Feel free to check that out.

langfuse/decorators/langfuse.py
```diff
@@ -0,0 +1,388 @@
...
+        Note:
+            - The decorator automatically manages observation IDs and context, but you can manually specify an observation ID using the `langfuse_observation_id` keyword argument when calling the wrapped function.
+            - To update observation or trace parameters (e.g., metadata, session_id), use `langfuse.update_current_observation` and `langfuse.set_current_trace_params` within the wrapped function.
+        """
+
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            langfuse = self._get_langfuse()
+            stack = observation_stack_context.get().copy()
+            parent = stack[-1] if stack else None
```
maxdeichmann, line 82: Quick question: say i have a function and annotate it with a trace. Within that function i start two threads, which execute two annotated functions. Will both get the same parent?

langfuse/decorators/langfuse.py
```diff
@@ -0,0 +1,388 @@
...
+                    observation.update(**observation_params)
+
+                # Remove observation from top of stack by resetting to initial stack
+                observation_stack_context.set(stack)
+
+            return result
+
+        return wrapper
+
+    def get_current_llama_index_handler(self):
```
maxdeichmann, line 131: This is great DX!!!!

langfuse/decorators/langfuse.py
```diff
@@ -0,0 +1,536 @@
...
+            - The decorator automatically manages observation IDs and context, but you can manually specify an observation ID using the `langfuse_observation_id` keyword argument when calling the wrapped function.
+            - To update observation or trace parameters (e.g., metadata, session_id), use `langfuse.update_current_observation` and `langfuse.set_current_trace_params` within the wrapped function.
+        """
+
+        def decorator(func: Callable) -> Callable:
+            return (
+                self.async_trace(func, as_type=as_type)
+                if asyncio.iscoroutinefunction(func)
+                else self.sync_trace(func, as_type=as_type)
+            )
```
maxdeichmann, line 113: Good implementation above, i think this is the way to go!

tests/test_decorators.py
```diff
@@ -0,0 +1,497 @@
...
+        lvl += 1
+
+    assert lvl == 3
+
+    assert llama_index_root_span is not None
+    assert any([o.name == "ChatPromptTemplate" for o in trace_data.observations])
+
+
+@pytest.mark.asyncio
+async def test_asyncio_concurrency_inside_nested_span():
```
maxdeichmann, line 380: Nice!

maxdeichmann (approved):

maxdeichmann (approved):

maxdeichmann (approved):

langfuse/decorators/langfuse_decorator.py
```diff
@@ -477,7 +485,7 @@ def get_current_langchain_handler(self):
 
         return observation.get_langchain_handler()
 
-    def get_current_trace_id(self):
+    def get_current_trace_id(self, log_warnings: bool = True):
```
maxdeichmann, line 488: Wdyt about always logging a warning here? I think this would be unexpected for the user to not get an id here.

maxdeichmann: I see the reason for the change below. In case of the openai integration, we do not want to log error messages. What do you think about adding the logic of these functions to the openai integration as well? 
Reasoning:
- these functions here are fairly shallow
- If we add log_warnings to the public interface, we need to explain them and cannot just remove them. Hence i would only expose what is necessary.

langfuse/utils/langfuse_singleton.py
```diff
@@ -17,14 +17,28 @@ def __new__(cls):
...
-            self._langfuse = Langfuse()
+            self._langfuse = Langfuse(
+                public_key=public_key,
+                secret_key=secret_key,
+                host=host,
+                debug=debug,
+                sdk_integration=sdk_integration,
+            )
 
+            print("LangfuseSingleton: Langfuse instance created.")
```
maxdeichmann, line 43: can you remove this?

PR #460, opened 2024-03-05
fix: fix proto serialization

SinghCoder:
Possible fixes https://github.com/langfuse/langfuse/issues/1327


langfuse/serializer.py
```diff
@@ -1,13 +1,16 @@
...
-from dataclasses import is_dataclass, asdict
 from json import JSONEncoder
 from typing import Any
 from uuid import UUID
 
-from langfuse.api.core import serialize_datetime
-
+from google.protobuf.json_format import MessageToJson
+from google.protobuf.message import Message
 from pydantic import BaseModel
```
SinghCoder, line 10: the re-ordering is done by [`Black`](https://pypi.org/project/black/) formatter, and since you don't have a CONTRIBUTING guide into this repo, so not sure which standard you follow.

langfuse/serializer.py
SinghCoder, line 1: NOTE: Tests will be added once I can get a review on whether the logic looks correct. As described in the issue, the second if looks dicey to me, took GPT-4 help to come up with this solution.

SinghCoder:
@marcklingen @maxdeichmann can I get a review on this please?

maxdeichmann:
Hi @SinghCoder thanks for raising this PR! I just checked it and with this we would add another dependency to our SDK. Is there a way to parse the proto objects without adding the google dependency?

SinghCoder:
> Hi @SinghCoder thanks for raising this PR! I just checked it and with this we would add another dependency to our SDK. Is there a way to parse the proto objects without adding the google dependency?

Hey @maxdeichmann, but we would need to atleast add the check that incoming object to serializer is a proto right? I mean even if I don't use `MessageToJson`, I do need the `isinstance(obj, Message)` no?

SinghCoder:
@maxdeichmann @marcklingen any more thoughts on this?

langfuse/serializer.py
```diff
@@ -29,6 +30,8 @@ def default(self, obj: Any):
 
         if is_dataclass(obj):
             return asdict(obj)
+        if isinstance(obj, Sequence) and not isinstance(obj, (str, bytes)):
```
maxdeichmann, line 33: In general, i like this approach.
Should this condition come below line 48 `isinstance(obj, (dict, list, str, int, float, type(None)))`?
Also, is it possible that obj is a Sequence and a str at the same time? If no, i think the second condition is not necessary, right?

SinghCoder: if I move it below L48, then yes we might not need it.

SinghCoder: I moved it below now @maxdeichmann

SinghCoder:
Hey @maxdeichmann @marcklingen, I would really like to close the PR soon. Can you please help expedite the review? Right now I am disabling publishing stuff to Langfuse in my code because of the errors being received.
I appreciate your time, and understand you might be busy with other priorities, but can we try to close this one soon please?

SinghCoder:
@maxdeichmann @marcklingen anything here?

PR #476, opened 2024-03-09
fix: stream support in query engines

hassiebp:
Fixes stream support in query engines. Addresses user issue: https://github.com/orgs/langfuse/discussions/828#discussioncomment-8725481

### DIFF
--- a/langfuse/task_manager.py
+++ b/langfuse/task_manager.py
@@ -109,6 +109,7 @@ class Consumer(threading.Thread):
                         "Item exceeds size limit (size: %s), dropping item.",
                         item_size,
                     )
+                    self._queue.task_done()
                     continue
                 items.append(item)
                 total_size += item_size
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "langfuse"
-version = "2.20.0"
+version = "2.20.1"
 description = "A client library for accessing langfuse"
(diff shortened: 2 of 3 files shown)


jimgoo:
Does the test you've implemented print the response? For my example below, it doesn't print the response at the end. Not sure if that's b/c the streaming generator has already yielding everything in the `LlamaIndexCallbackHandler` when it gets converted to a string.

```python
from llama_index.core import Settings, SimpleDirectoryReader, VectorStoreIndex, StorageContext, load_index_from_storage
from llama_index.core.callbacks import CallbackManager

from langfuse.llama_index import LlamaIndexCallbackHandler

langfuse_callback_handler = LlamaIndexCallbackHandler(
    host="http://localhost:3001",
    debug=True,
)

Settings.callback_manager = CallbackManager([langfuse_callback_handler])

storage_dir = '/Users/jimgoo/git/rcp/trase-demo/llama-index/backend/storage'
storage_context = StorageContext.from_defaults(persist_dir=storage_dir)
index = load_index_from_storage(storage_context, chunk_size=256, show_progress=True)

use_streaming = True
chat_engine = index.as_query_engine(streaming=use_streaming)
validated_prompt = 'Hey how are you?'
response = chat_engine.query(validated_prompt)

# nothing prints here
print('-'*100)
for token in response.response_gen:
    print(token, end="")
```

maxdeichmann (approved):

PR #509, opened 2024-03-27
Replacing root logging calls with langfuse logger

samyxdev:
I found two instances where the root logger from logging was being called, hence making impossible to disable logging from Langfuse externally (ie. in a script using the langfuse package).

### DIFF
--- a/langfuse/client.py
+++ b/langfuse/client.py
@@ -2234,6 +2234,8 @@ class DatasetItemClient:
             run_name (str): The name of the dataset run.
         """
         observation_id = None
+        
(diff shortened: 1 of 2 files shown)


samyxdev:
Hey @marcklingen, would you have any feedback on the PR ? It would be much appreciated :)

marcklingen:
Thanks for the ping, @maxdeichmann will have a look

maxdeichmann (approved):

marcklingen:
Thanks again for your contribution @samyxdev 🙏

PR #518, opened 2024-03-31
Fix: langchain async and sync batch calls

noble-varghese:
# Description
Test Langchain's synchronous, asynchronous streaming, batch, and invoke methods, and resolve the asynchronous issue encountered with Langchain batch calls.

Key Modifications:
- Users encountered issues due to parallel processing in Langchain integrations, resulting in missing "unable to find run-id" error messages in the logs.
- Integration tests have been implemented for both synchronous and asynchronous streaming methods.


langfuse/callback/langchain.py
```diff
@@ -743,6 +743,9 @@ def _update_trace(self, run_id: str, parent_run_id: Optional[str], output: any):
         ):
             self.trace = self.trace.update(output=output)
 
+        # Remove the run details after updating the trace.
```
maxdeichmann, line 746: Can we quickly rename the function? E.g. _update_trace_and_remove_state(...) to make this more readable?

maxdeichmann: If the tests complete, i think we are good to go here. Can you add assertions to e.g. https://github.com/noble-varghese/langfuse-python/blob/c1aee1422c48c7cabb4fc4febf7407fca11ccfa4/tests/test_langchain.py#L886 to check that the runs map is empty?

noble-varghese: Fixed on Issue #527 and added assertions as well.

PR #532, opened 2024-04-04
fix: adding trace i/o in langfuse openai integration

noble-varghese:
# Description
The LangFuse OpenAI integration previously failed to include the input and output data in the trace body, limiting it only to the generation.

Key Changes:
- Implemented sending input/output in the trace body.
- Trace input/output is **not** updated when a trace_id or observation_id is provided in the inputs. This prevents inadvertent updates to nested traces.


langfuse/openai.py
```diff
@@ -325,6 +330,8 @@ async def _get_langfuse_data_from_async_streaming_response(
         resource, responses
     )
 
+    langfuse.trace(id=generation.trace_id, output=completion)
```
maxdeichmann, line 333: In these cases, we do not know anymore whether the trace was generated by the user becore and whether he provided the traceId himself. Can we somehow get that information here and only update in this case? Similar to how we do that with the input.

langfuse/openai.py
```diff
@@ -505,7 +530,7 @@ async def _wrap_async(
     start_time = _get_timestamp()
     arg_extractor = OpenAiArgsExtractor(*args, **kwargs)
 
-    generation = _get_langfuse_data_from_kwargs(
+    generation, is_nested_trace = _get_langfuse_data_from_kwargs(
```
noble-varghese, line 533: Was wondering if we could return a nested function bool from the existing function which would make the implementation easier to review and also possibly reduce the touch points in the code. If we change, we change on ly inside the function and rest should fall in place.

maxdeichmann:
Implementation looks good to me. Could you add a test to the openai test function ensuring we only add io in the case discussed?

noble-varghese:
@maxdeichmann I have added a test for checking the scenario we discussed where the user provides a trace_id in the openai integration as well as updated other tests to verify the trace i/o to be validated again input and out to and from model.

maxdeichmann (approved):

PR #538, opened 2024-04-07
Fix: Fixes empty input in the langchain stream calls

noble-varghese:
# Description
- Fixes the input in the stream calls by adding the inputs in streaming and astream calling

Issue: https://github.com/langfuse/langfuse/issues/1589


noble-varghese:
@hassiebp Yes, in the case of the streaming data the input passed on the `on_chain_start` method from langchain contains a dummy input and not the actual input passed by the user. The complete input is only delivered at the end of the chain on the `on_chain_end` method. 
You can check the code [here](https://github.com/langchain-ai/langchain/blob/ba9e0d76c1691608c3ecc8510db1f53c7cb36dcd/libs/core/langchain_core/runnables/base.py#L1862C1-L1864C27) from langchain where the input is being passed as a dummy input for the streaming data.

PR #543, opened 2024-04-08
fix: disable HTML escaping of prompt compile inputs

hassiebp:
Disable HTML escaping of prompt compile inputs

**Changes**
- Removes `chevron` as template-replacement library as it does not support disabling HTML escaping
- Introduced custom `compile_template_string` function with test suite
- Jinja2 has the following restrictions on [variable names](https://jinja.palletsprojects.com/en/3.1.x/api/#notes-on-identifiers). If the prompt feature users decide to keep a variable unused and see it rendered within double curly braces, Jinja2 would throw an error as it cannot parse the variable name.

### DIFF
--- a/langfuse/client.py
+++ b/langfuse/client.py
@@ -2317,6 +2317,8 @@ class DatasetItemClient:
             run_name (str): The name of the dataset run.
         """
         observation_id = None
+        
+        log = logging.getLogger("langfuse")
 
         if isinstance(observation, StatefulClient):
             # flush the queue before creating the dataset run item
@@ -2331,7 +2333,7 @@ class DatasetItemClient:
                 "observation parameter must be either a StatefulClient or a string"
             )
 
-        logging.debug(
+        log.debug(
             f"Creating dataset run item: {run_name} {self.id} {observation_id}"
         )
         self.langfuse.client.dataset_run_items.create(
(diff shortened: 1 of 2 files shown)


maxdeichmann (approved):

langfuse/model.py
```diff
@@ -73,17 +72,56 @@ def get_langchain_prompt(self):
         pass
 
     @staticmethod
-    def get_langchain_prompt_string(content: str):
+    def _get_langchain_prompt_string(content: str):
         return re.sub(r"\{\{(.*?)\}\}", r"{\g<1>}", content)
 
+    @staticmethod
+    def _compile_template_string(content: str, **kwargs) -> str:
```
maxdeichmann, line 79: Did you try to implement that with regex? I would prefer that over looping ourselves through the string.

langfuse/utils/compile_template_string.py
```diff
@@ -0,0 +1,37 @@
...
+
+        if var_start == -1:
+            result_list.append(content[curr_idx:])
+            break
+
+        # Find the next closing tag
+        var_end = content.find(closing, var_start)
+
+        if var_end == -1:
+            result_list.append(content[curr_idx:])
```
maxdeichmann, line 20: This here would happen in case of a male formatted sting, right? `{{var1}} something{{random {{var2}}` would not match var2, right?

PR #556, opened 2024-04-10
fix(prompt): compile None value to empty string

fancyweb:
Thanks to https://github.com/langfuse/langfuse-python/pull/543 I can now get rid of my own variable interpolation system that circumvented HTML escaping issues :partying_face:

However, one behavior that I think we should preserve is compiling `None` values to `""` instead of `"None"` (as chevron did).


marcklingen:
Released: https://github.com/langfuse/langfuse-python/releases/tag/v2.23.2

PR #559, opened 2024-04-10
fix(prompt): allow "content" prompt var again

fancyweb:
One more issue after https://github.com/langfuse/langfuse-python/pull/543 is that having an existing `content` prompt var doesn't work anymore:
```
TypeError: BasePromptClient._compile_template_string() got multiple values for argument 'content'
```
Let's reallow it :pray:

### DIFF
--- a/langfuse/model.py
+++ b/langfuse/model.py
@@ -76,7 +76,8 @@ class BasePromptClient(ABC):
         return re.sub(r"\{\{(.*?)\}\}", r"{\g<1>}", content)
 
     @staticmethod
-    def _compile_template_string(content: str, **kwargs) -> str:
+    def _compile_template_string(*args, **kwargs) -> str:
+        content = args[0]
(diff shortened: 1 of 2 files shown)


langfuse/model.py
```diff
@@ -76,7 +76,8 @@ def _get_langchain_prompt_string(content: str):
         return re.sub(r"\{\{(.*?)\}\}", r"{\g<1>}", content)
 
     @staticmethod
-    def _compile_template_string(content: str, **kwargs) -> str:
+    def _compile_template_string(*args, **kwargs) -> str:
```
fancyweb, line 79: Not sure at all about the way I fixed this since I'm new to Python, I let you handle it if there's a better way :smiley:

marcklingen:
Released: https://github.com/langfuse/langfuse-python/releases/tag/v2.23.2

PR #597, opened 2024-04-18
fix: uriencode dataset names in get_dataset(name)

marcklingen:


hassiebp:
@marcklingen is this still an issue / are we planning to merge this? If not, let's close as stale

PR #598, opened 2024-04-18
Bugfix: to fix issue #599 where `meta` can be `None`

zoltan-fedor:
`meta` can be `None`, see https://github.com/langfuse/langfuse-python/blob/main/langfuse/callback/langchain.py#L722, but `meta.update()` at https://github.com/langfuse/langfuse-python/blob/main/langfuse/callback/langchain.py#L445 would fail on `None`

This fixes bug report https://github.com/langfuse/langfuse-python/issues/599


marcklingen (approved):
thanks!

PR #605, opened 2024-04-19
Fix span input empty when call langchain stream

wodadehencou:
https://github.com/langfuse/langfuse/issues/1589#issuecomment-2066115293


maxdeichmann:
@wodadehencou thanks a lot for the contribution. I just released this here: https://github.com/langfuse/langfuse-python/releases/tag/v2.29.0

PR #622, opened 2024-04-26
Adding observation_name as argument of observe()

samyxdev:
Hi,

I propose a change in the `observe()` decorator in order to, optionally set the observation name from the decorator instead of having to call `langfuse_context.update_current_trace` to do so.

What are your thoughts ?


marcklingen:
Looks good to me, cc @hassiebp who built the decorator to review

I'd probably use `name` instead of `observation_name` to keep things simple.

samyxdev:
Amazing, thanks @hassiebp and @marcklingen !

PR #633, opened 2024-04-30
feat: adds labels to prompts

hassiebp:
Adds support for prompt labels

### DIFF
--- a/langfuse/client.py
+++ b/langfuse/client.py
@@ -26,12 +26,12 @@ from langfuse.api.resources.observations.types.observations_views import (
 from langfuse.api.resources.prompts.types import (
     CreatePromptRequest_Chat,
     CreatePromptRequest_Text,
-    ChatMessage,
 )
 from langfuse.model import (
     CreateDatasetItemRequest,
     CreateDatasetRequest,
     CreateDatasetRunItemRequest,
+    ChatMessageDict,
     DatasetItem,
     DatasetRun,
     DatasetStatus,
@@ -555,9 +555,9 @@ class Langfuse(object):
 
         Args:
             name (str): The name of the prompt to retrieve.
-            version (Optional[int]): The version of the prompt. If not specified, the `active` version is returned. Specify either version or label, not both.
 
         Keyword Args:
+            version (Optional[int]): The version of the prompt to retrieve. If no label and version is specified, the `production` label is returned. Specify either version or label, not both.
             label: Optional[str]: The label of the prompt to retrieve. If no label and version is specified, the `production` label is returned. Specify either version or label, not both.
             cache_ttl_seconds: Optional[int]: Time-to-live in seconds for caching the prompt. Must be specified as a
             keyword argument. If not set, defaults to 60 seconds.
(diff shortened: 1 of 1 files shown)


langfuse/model.py
```diff
@@ -151,11 +158,13 @@ def get_langchain_prompt(self):
 class ChatPromptClient(BasePromptClient):
     def __init__(self, prompt: Prompt_Chat):
         super().__init__(prompt)
-        self.prompt = prompt.prompt
+        self.prompt = [
+            ChatMessageDict(role=p.role, content=p.content) for p in prompt.prompt
+        ]
 
-    def compile(self, **kwargs) -> List[ChatMessage]:
+    def compile(self, **kwargs) -> List[ChatMessageDict]:
```
maxdeichmann, line 165: Is this a breaking change if someone imported `ChatMessage` originally? I think yes. The import path would have involved a path to the api module, correct?
If yes, this is hard to change, if no, i would suggest to keep the naming of the exported package.

langfuse/prompt_cache.py
```diff
@@ -38,5 +38,19 @@ def set(self, key: str, value: PromptClient, ttl_seconds: Optional[int]):
...
+
+        if version is not None:
+            parts.append(f"version:{version}")
+
+        elif label is not None:
+            parts.append(f"label:{label}")
+
+        else:
+            # Default to production labeled prompt
+            parts.append("label:production")
```
maxdeichmann, line 54: this assumption makes sense.

tests/test_prompt.py
```diff
@@ -167,6 +209,38 @@ def test_create_prompt_with_null_config():
...
+    assert default_prompt_client.version == 2
+    assert default_prompt_client.prompt == "test prompt 2"
+    assert default_prompt_client.labels == ["production"]
+
+    first_prompt_client = langfuse.get_prompt(prompt_name, 1)
+    assert first_prompt_client.version == 1
+    assert first_prompt_client.prompt == "test prompt 1"
+    assert first_prompt_client.labels == []
+
+    second_prompt_client = langfuse.get_prompt(prompt_name, version=2)
```
maxdeichmann, line 233: image
Do you have the same signature when hovering?

langfuse/client.py
```diff
@@ -538,6 +542,7 @@ def get_prompt(
         name: str,
         version: Optional[int] = None,
         *,
+        label: Optional[str] = None,
```
maxdeichmann, line 545: would it make sense to add another overload like:


```
    @overload
    def get_prompt(
        self,
        name: str,
        *,
        version: Optional[int] = None,
        label: Optional[str] = None,
        type: Literal["text"] = "text",
        cache_ttl_seconds: Optional[int] = None,
    ) -> TextPromptClient: ...

```

tests/test_prompt.py
marcklingen: do we need `version` as `arg` for backwards compatibility while all others are `kwargs`?

langfuse/client.py
```diff
@@ -648,7 +666,8 @@ def create_prompt(
         *,
         name: str,
         prompt: Union[str, List[ChatMessage]],
-        is_active: bool,
+        is_active: Optional[bool] = None,  # deprecated
+        labels: List[str] = [],
```
maxdeichmann, line 670: as these are all named arguments, order does not matter - correct?

langfuse/prompt_cache.py
marcklingen: agree, lgtm

marcklingen (approved):
looks good to me, awesome!

PR #644, opened 2024-05-03
feat(langchain): adds update_parent

hassiebp:
Adds `update_parent` argument to `LangchainCallbackHandler`. Setting it to true will update provided parent trace or observation with the results of the langchain run.

### DIFF
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -300,7 +300,7 @@ def test_langchain_dataset():
     assert sorted_observations[0].id == sorted_observations[1].parent_observation_id
     assert sorted_observations[0].parent_observation_id is None
 
-    assert trace.name == "dataset-run"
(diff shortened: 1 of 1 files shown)


langfuse/client.py
```diff
@@ -2251,12 +2253,15 @@ def update(
...
 
-    def get_langchain_handler(self):
+    def get_langchain_handler(self, update_parent: bool = False):
         """Get langchain callback handler associated with the current trace.
 
         This method creates and returns a CallbackHandler instance, linking it with the current
         trace. Use this if you want to group multiple Langchain runs within a single trace.
 
+        Args:
+            update_parent (bool): If set to True, the parent trace or span will be updated with the outcome of the Langchain run.
```
marcklingen, line 2263:
```suggestion
            update_parent (bool): If set to True, the parent trace or observation will be updated with the outcome of the Langchain run.
```

PR #651, opened 2024-05-05
fix: do not add tokens to cost calculation in case of openai errors

marcklingen:


marcklingen:
Todo: also add on async wrapper

maxdeichmann (approved):

PR #663, opened 2024-05-07
fix(decorator): serialize arbitrary decorated function inputs

hassiebp:
Attempt decorated function serialization right when capturing the input. If object is not serializable, log the type name.


maxdeichmann (commented):
added small comment. good to go

PR #665, opened 2024-05-07
fix(langchain): support llamacpp

maxdeichmann:


omarcevi:
Hello, it seems that with this commit langfuse still wont parse LlamaCpp models because `invocations` and `model_path` are not in `serialized` but rather in `kwargs`
so in models_by_id list LlamaCpp should be defined like this:
`("LlamaCpp", ["invocation_params", "model_path"], "kwargs")`

maxdeichmann:
thanks @omarcevi i fixed it [here](https://github.com/langfuse/langfuse-python/commit/31b539262167428192a30e90f9f7957371a4041c#commitcomment-141763970)

PR #669, opened 2024-05-07
feat(core): adds httpx client and 'enabled'-switch for Langfuse

hassiebp:
- Adds `enabled` switch that decides whether to send observability events to Langfuse (langfuse.trace/generation/event/score/span)
- Adds passing a custom HTTPX client for `llama-index` and `langchain` integrations

### DIFF
--- a/langfuse/client.py
+++ b/langfuse/client.py
@@ -106,6 +106,7 @@ class Langfuse(object):
         timeout: int = 20,  # seconds
         sdk_integration: Optional[str] = "default",
         httpx_client: Optional[httpx.Client] = None,
+        enabled: Optional[bool] = True,
     ):
         """Initialize the Langfuse client.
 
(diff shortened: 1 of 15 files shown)


langfuse/client.py
```diff
@@ -163,15 +171,15 @@ def __init__(
...
+            self.log.warning(
+                "Langfuse client is disabled since no public_key was provided as a parameter or environment variable 'LANGFUSE_PUBLIC_KEY'."
             )
 
         if not secret_key:
-            self.log.warning("secret_key is not set.")
-            raise ValueError(
-                "secret_key is required, set as parameter or environment variable 'LANGFUSE_SECRET_KEY'"
+            self.enabled = False
+            self.log.warning(
```
marcklingen, line 181: dx improvement: log only a single warning when disabled and state reason in this warning. currently user gets up to 3 warnings

langfuse/task_manager.py
```diff
@@ -251,6 +254,9 @@ def init_resources(self):
             self._consumers.append(consumer)
 
     def add_task(self, event: dict):
+        if not self._enabled:
```
marcklingen, line 257: this is so clean 🤩

langfuse/client.py
```diff
@@ -140,6 +142,27 @@ def __init__(
...
+        elif not public_key:
+            self.enabled = False
+            self.log.warning(
+                "Langfuse client is disabled since no public_key was provided as a parameter or environment variable 'LANGFUSE_PUBLIC_KEY'."
+            )
+
+        elif not secret_key:
+            self.enabled = False
+            self.log.warning(
+                "Langfuse client is disabled since no secret_key was provided as a parameter or environment variable 'LANGFUSE_SECRET_KEY'."
```
maxdeichmann, line 163: can we add a link to the docs here on how to set up?

PR #670, opened 2024-05-07
fix: add model support for langchain_aws

bell-steven:
This adds model parsing support for ChatBedrock from `langchain_aws`, since BedrockChat from has been deprecated.

```
 The class `BedrockChat` was deprecated in LangChain 0.0.34 and will be removed in 0.3. An updated version of the class exists in the langchain-aws package and should be used instead. To use it run `pip install -U langchain-aws` and import as `from langchain_aws import ChatBedrock`
```


maxdeichmann (approved):

maxdeichmann:
thank you @bell-steven for the contribution!

PR #691, opened 2024-05-19
Fix incorrect types in client os.getenv calls

rohan-mehta:
When I set `LANGFUSE_FLUSH_AT=50`, I saw this error in logs:
```
File "/usr/local/lib/python3.11/site-packages/langfuse/task_manager.py", line 99, in _next
while len(items) < self._flush_at:
^^^^^^^^^^^^^^^^^^^^^^^^^^^
 TypeError: '<' not supported between instances of 'int' and 'str'
```

This is because env vars are all strings:
```
$ FOO=50 python -c 'import os; print(type(os.getenv("FOO")))'
<class 'str'>
```

This fixes the code to instead get with a default value, and then cast to the expected type.

### DIFF
--- a/langfuse/client.py
+++ b/langfuse/client.py
@@ -146,13 +146,15 @@ class Langfuse(object):
         public_key = public_key or os.environ.get("LANGFUSE_PUBLIC_KEY")
         secret_key = secret_key or os.environ.get("LANGFUSE_SECRET_KEY")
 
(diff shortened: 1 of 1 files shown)


maxdeichmann:
@rohan-mehta thanks a lot for the contribution and fixing this. We released this in: https://github.com/langfuse/langfuse-python/releases/tag/v2.33.0

PR #714, opened 2024-05-30
fix(openai): allow for contextmanager in wrapped generators

hassiebp:
Allow for consumption of returned streaming generators inside a  contextmanager.


langfuse/openai.py
```diff
@@ -648,3 +602,97 @@ def _filter_image_data(messages: List[dict]):
...
+        # Avoiding the trace-update if trace-id is provided by user.
+        if not self.is_nested_trace:
+            self.langfuse.trace(id=self.generation.trace_id, output=completion)
+
+        _create_langfuse_update(
+            completion, self.generation, completion_start_time, model=model
+        )
+
+
+class LangfuseResponseGeneratorAsync:
```
maxdeichmann, line 654: what about typing this as the openai sdk does it? We could subclass from `AsyncResponseContextManager`

PR #767, opened 2024-06-28
feat(scores): add `configId`, `dataType`, `stringValue` to python SDK

marliessophie:


tests/test_prompt.py
marliessophie, line 1: Automatic lint changes only

PR #776, opened 2024-07-03
perf: for large payloads drop fields by size

hassiebp:
If events to be delivered have too large a payload, drop input / output / metadata fields from the largest first until size limit is met.


maxdeichmann:
Why are we tackling it this way? What was the issue before? Implementation looks solid to me, but we will have non deterministic behavior now. In some cases, a certain field will be dropped than in other cases depending on the size of others. I would be fine with that.

PR #816, opened 2024-07-23
Langchain prompt single braces

JannisConen:
**Goal:** be able to use single braces in langfuse prompts later interacting with langchain

**Why?**
I am working a lot with generated latex which kind of revolves around { and }. Also, I am using langchain as a framework. Therefore I am not able to use { and } in my prompts on langfuse because then langchain expects them to hold variables inside them. To escape { and } for langchain, you would use {{ and }} but in langfuse {{ and }} circumenvent variables.

**My quick suggestion**
First replace all single { and } with temporary placeholders, then have the usual langfuse -> langchain prompt transformation and finally replace the temporary placeholders with double quotes.

I am a little unsure how this might impact the vast variety of prompts there is. Maybe some people use this as a feature to have variables in the prompt for langchain and not mark them variables for langfuse.

### DIFF
--- a/langfuse/model.py
+++ b/langfuse/model.py
@@ -83,7 +83,20 @@ class BasePromptClient(ABC):
 
     @staticmethod
     def _get_langchain_prompt_string(content: str):
-        return re.sub(r"{{\s*(\w+)\s*}}", r"{\g<1>}", content)
+        # replace single `{` with the temporary replacement content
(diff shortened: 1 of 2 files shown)


hassiebp:
Thanks again for raising this! Closing this as stale / no longer relevant in Python SDK v3. If you continue to see issues, please open a Github issue 👍🏾

PR #862, opened 2024-08-13
fix: Create py.typed

kongzii:
Hi, just a small fix: without this, mypy doesn't consider langfuse's typing.

Given the following test file

```
from langfuse.decorators import observe


@observe()
def test(x: int) -> str:
    return str(x)


y = test(5)
reveal_type(y)
```

with this change, mypy looks like this:

```
test.py:9: error: Argument 1 to "test" has incompatible type "str"; expected "int"  [arg-type]
test.py:10: note: Revealed type is "builtins.str"
```

without this change (with the latest langfuse version), it looks like this:

```
test.py:10: note: Revealed type is "Any"
Success: no issues found in 1 source file
```

---

I see there is some discussion already in https://github.com/langfuse/langfuse/issues/2169, but it got stalled.

It is true that if I just run `mypy` on this repository, there are many complaints:

```
main $ mypy . | wc -l                                                                                                                                                                                                                           [10:55:22]
     517
```

So it would require a big effort to fix the whole library for mypy, but on the other hand, it's a shame that currently, any project that uses `@observe` is losing type safety, which can be fixed by this change.

As a hot fix, without this change, having this wrapper in our code repository around the langfuse's observer seems like a way to go, but it's not a great solution:

```python
from langfuse.decorators.langfuse_decorator import ( 
    langfuse_context,
    observe as original_observe,
)
from typing import TypeVar, ParamSpec, Optional, Literal, Callable, Iterable, Any

P = ParamSpec("P")
R = TypeVar("R")


def observe(
    name: Optional[str] = None,
    as_type: Optional[Literal["generation"]] = None,
    capture_input: bool = True,
    capture_output: bool = True,
    transform_to_string: Optional[Callable[[Iterable[Any]], str]] = None,
) -> Callable[[Callable[P, R]], Callable[P, R]]:
    casted: Callable[[Callable[P, R]], Callable[P, R]] = original_observe(
        name=name,
        as_type=as_type,
        capture_input=capture_input,
        capture_output=capture_output,
        transform_to_string=transform_to_string,
    )
    return casted
```


hassiebp:
Thanks again for raising this! Closing this as stale / no longer relevant in Python SDK v3. If you continue to see issues, please open a Github issue 👍🏾

kongzii:
Created the issue https://github.com/langfuse/langfuse/issues/7880.

PR #874, opened 2024-08-21
Support for Watsonx ai model for langchain extension

barvhaim:
Changes for supporting Watsonx.ai (https://python.langchain.com/v0.2/docs/integrations/llms/ibm_watsonx/)

image

### DIFF
--- a/langfuse/Sampler.py
+++ b/langfuse/Sampler.py
@@ -26,6 +26,8 @@ class Sampler:
                 trace_id = event["body"]["trace_id"]
             elif "traceId" in event["body"]:
                 trace_id = event["body"]["traceId"]
+            elif "id" in event["body"]:
+                trace_id = event["body"]["id"]
             else:
                 log.error("Unexpected event format: No trace id found in event")
                 return True
--- a/langfuse/Sampler.py
+++ b/langfuse/Sampler.py
@@ -26,8 +26,6 @@ class Sampler:
                 trace_id = event["body"]["trace_id"]
             elif "traceId" in event["body"]:
                 trace_id = event["body"]["traceId"]
-            elif "id" in event["body"]:
-                trace_id = event["body"]["id"]
             else:
                 log.error("Unexpected event format: No trace id found in event")
                 return True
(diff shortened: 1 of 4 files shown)


langfuse/Sampler.py
```diff
@@ -26,6 +26,8 @@ def sample_event(self, event: dict):
                 trace_id = event["body"]["trace_id"]
             elif "traceId" in event["body"]:
                 trace_id = event["body"]["traceId"]
+            elif "id" in event["body"]:
+                trace_id = event["body"]["id"]
```
maxdeichmann, line 30: Why did you make this change? Above, we get the event.body.id, in case the event is a trace-create type.

langfuse/utils/__init__.py
```diff
@@ -67,8 +67,10 @@ def _convert_usage_input(usage: typing.Union[pydantic.BaseModel, ModelUsage]):
         for k in (
             "promptTokens",
             "prompt_tokens",
+            "input_token_count",
             "completionTokens",
             "completion_tokens",
+            "generated_token_count",
```
maxdeichmann, line 73: Can we revert this change and handle this in the Langchain callback handler?

- Functionality in here should be abstracted for Langfuse. Hence, everything going in here should already have been converted to prompt_tokens, completion_tokens etc.

How does this work in the langchain case for IBM? We parse the usage from all the different models [here](https://github.com/langfuse/langfuse-python/blob/4d7112a4b293bff5a329a4351abfefbc468d87b0/langfuse/callback/langchain.py#L731). Could you add your logic to `_parse_usage`?

Thanks

langfuse/Sampler.py
barvhaim: @maxdeichmann I inspected the event object and I saw my chains didn't have traceId, just an id so I added it as the traceId.. it was showing a warning about it

langfuse/utils/__init__.py
barvhaim: sure I'll try to make the change

langfuse/Sampler.py
barvhaim: Ok I removed it and it seems fine, I'll commit soon

langfuse/utils/__init__.py
barvhaim: usage for us looks like - 
image

barvhaim: @maxdeichmann I fixed and pushed

PR #878, opened 2024-08-22
Anthropic prompt caching

mrdrprofuroboros:


hey-nicolasklein:
> Thanks for your contribution! We are working on more generalized support for different cost tracking scenarios on the server side. Closing this PR in favor of that approach.

Is there an open Issue for this? Would also love to have accurate cost tracking for [Anthropic Claude PDF Support](https://docs.anthropic.com/en/docs/build-with-claude/pdf-support#calculate-expected-token-usage). Which combines cost for image-based and text-based calculations.

PR #889, opened 2024-08-29
The LangchainCallbackHandler supports statistical delay of the first …

Hacky-DH:
In the on_llm_new_token method of the LangchainCallbackHandler class, update the completion_start_time parameter of the generation using the current time. This way, you will be able to see the value of Time to First Token on the front-end page.

I issue here https://github.com/langfuse/langfuse/issues/3130


Hacky-DH:
merged here https://github.com/langfuse/langfuse-python/pull/938, close this

PR #893, opened 2024-08-29
Added support to get usage for ChatGoogleGenerativeAI

NeerajG03:
This PR has been raised to add support to track `ChatGoogleGenerativeAI`'s langchain call usage.
A simple check for usage_metadata attribute and then parsing using the  `_parse_usage_model`


maxdeichmann:
@NeerajG03 did you test this locally?

NeerajG03:
I'm sorry i'm very new to contributing for projects. 
If you are asking if i have run this test `poetry run pytest -s -v --log-cli-level=INFO`, then thats a yes. 👍

NeerajG03:
**Here is the screenshot of the run summary**
Screenshot 2024-08-29 at 10 27 56 PM

aunitt:
This also worked for me for ChatVertexAI.

It appears to fix my issues which appear to be those mentioned in https://github.com/langfuse/langfuse/issues/5468

langfuse/callback/langchain.py
```diff
@@ -1139,6 +1139,11 @@ def _parse_usage(response: LLMResult):
                     break
```
ReneFabricius, line 1139: I would suggest also adding the following:

```suggestion
                    if llm_usage:
                        break
```

as in my testing with ChatVertexAI gemini-2.0-flash `generation_info` contains empty `usage_metadata` while `usage_metadata` in `message` are valid.

ReneFabricius, line 1139: I would suggest adding
```suggestion
                    if llm_usage:
                        break
```

or even better modify the previous condition to:

```
                if generation_chunk.generation_info and (
                        generation_chunk.generation_info.get("usage_metadata", None)
                    ):
```

as in my testing with ChatVertexAI `generation_info` contains empty `usage_metadata` while `message` contains valid one.

langfuse/callback/langchain.py
```diff
@@ -1139,6 +1139,11 @@ def _parse_usage(response: LLMResult):
...
 
                 message_chunk = getattr(generation_chunk, "message", {})
+
+                if hasattr(message_chunk, "usage_metadata") and message_chunk.usage_metadata is not None: #for ChatGoogleGenerativeAI
+                    llm_usage = _parse_usage_model(message_chunk.usage_metadata)
+                    break
+
```
ReneFabricius, line 1146: Looking at it better, this should have no effect, as a few lines down, there is:

```
or getattr(message_chunk, "usage_metadata", None) # for Ollama
```

PR #903, opened 2024-09-03
Updated deprecated logging warning in langfuse_decorator.py

demux79:
Updated deprecated logging warning

langfuse_decorator.py:788: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
  self._log.warn("No observation found in the current context")


marcklingen:
thank you!

PR #907, opened 2024-09-04
fix llamaindex completion serialize error

nightosong:
While I use langfuse(2.43.3) with llamaindex(0.11.4), I get an error output:
```
received error response: {'message': 'Invalid request data', 'errors': ['Expected object, received string']}
Received 400 error by Langfuse server, not retrying: {'message': 'Invalid request data', 'errors': ['Expected object, received string']}
```
Then I debug the task_manager file and find this code `data = json.dumps(kwargs, cls=EventSerializer)` .
In my case, the output object in body is type of `<class 'llama_index.core.base.llms.types.CompletionResponse'>`.
When it executes this line of  `default()` in EventSerializer, a TypeError will be thrown.
TypeError: `'MockValSer' object cannot be converted to 'SchemaSerializer'` (pydantic 2.8.2)
```
if isinstance(obj, BaseModel):
    return obj.dict()
```
So, I force to convert the completion response to str type. I am not sure if it is a good solution.


hassiebp:
Thanks for raising this @nightosong ! The issue is addressed in https://github.com/langfuse/langfuse-python/pull/916 . Could you please retry after merging and releasing that fix and open a fresh issue if that persists?

nightosong:
> Thanks for raising this @nightosong ! The issue is addressed in #916 . Could you please retry after merging and releasing that fix and open a fresh issue if that persists?

thanks, it works

PR #925, opened 2024-09-13
fix(langchain) python langchain retriever - parent run not found 

Meteord:
if an langchain retriever is without a chain, no parent run can be found:
```
retriever =  vectorstore.as_retriever(...)
retriever.invoke(input=query, config={"callbacks": [langfuse_handler]})
```

In this case, no exception is thrown but a new parent trace generated.

fixes https://github.com/langfuse/langfuse/issues/3223


hassiebp (approved):

hassiebp:
Thanks a lot for your contribution @Meteord ! 🙏🏾

PR #945, opened 2024-10-03
fix(serializer): Fix None serialization without langchain installed

abrasseu:
This PR aims to fix the serialization of `None` to `"<not serializable object of type: NoneType>"` when langchain is not installed.
Related issue: https://github.com/langfuse/langfuse/issues/3570


hassiebp:
Thanks for your contribution @abrasseu ! 🙏🏾

PR #974, opened 2024-10-22
fix: replace deprecated dict references

MarcTatam:
Should close https://github.com/langfuse/langfuse/issues/2946

All other usages of .dict() appeared to be with a V1 pydantic model, so should not be changed until a v2 model is used.

----

> [!IMPORTANT]
> Replace deprecated `dict()` with `model_dump()` in `task_manager.py` and `utils/__init__.py` for Pydantic v2 compatibility.
> 
>   - **Behavior**:
>     - Replace deprecated `dict()` with `model_dump()` in `task_manager.py` and `utils/__init__.py` for Pydantic v2 compatibility.
>     - Does not change other `dict()` usages with Pydantic v1 models.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse-python&utm_source=github&utm_medium=referral)<sup> for 1fb0f060be351efbc7b35f23790db6cec76414a8. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/langfuse/task_manager.py
+++ b/langfuse/task_manager.py
@@ -223,7 +223,7 @@ class Consumer(threading.Thread):
             sdk_name=self._sdk_name,
             sdk_version=self._sdk_version,
             public_key=self._public_key,
-        ).dict()
+        ).model_dump()
 
         @backoff.on_exception(
             backoff.expo, Exception, max_tries=self._max_retries, logger=None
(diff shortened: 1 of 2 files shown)


hassiebp:
@MarcTatam Thanks for your contribution! We have to make sure to support users on Pydantic V1 as well, and for them `model_dump` would fail. It could be worth refactoring the serialization into a small util that checks the pydantic version and calls `dict` or `model_dump` depending on the installed Pydantic version.

MarcTatam:
Understood I'll close this and open a new one later.

PR #978, opened 2024-10-25
fix: 'NoneType' object has no attribute '__dict__'

bettlebrox:
Handle case when usage is None.

https://github.com/langfuse/langfuse/issues/3757


hassiebp:
Thanks for your contribution @bettlebrox !

PR #1010, opened 2024-11-26
fix: correctly track falsy output values

tux-type:
Fixes [issue 4439](https://github.com/langfuse/langfuse/issues/4439) where falsy (`False`, `0`, etc.) values get tracked as `null`.

----

> [!IMPORTANT]
> Fixes issue in `langfuse_decorator.py` where falsy values were incorrectly tracked as `null`.
> 
>   - **Behavior**:
>     - Fixes issue where falsy values (`False`, `0`, etc.) were incorrectly tracked as `null` in `_handle_call_result()` in `langfuse_decorator.py`.
>     - Now checks `result is not None` instead of `result` to correctly handle falsy values.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse-python&utm_source=github&utm_medium=referral)<sup> for 685271adca003e1edbf5dc9b258e4ab5717f2337. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/langfuse/decorators/langfuse_decorator.py
+++ b/langfuse/decorators/langfuse_decorator.py
@@ -452,7 +452,7 @@ class LangfuseDecorator:
                 # Objects are later serialized again so deserialization is necessary here to avoid unnecessary escaping of quotes.
                 json.loads(
                     json.dumps(
-                        result if result and capture_output else None,
+                        result if result is not None and capture_output else None,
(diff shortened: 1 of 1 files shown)


marcklingen:
thank you!

hassiebp (approved):

PR #1013, opened 2024-11-27
Add variable name extraction methods to prompt clients

rhighs:
### Overview
PR introduces a `_find_variable_names` static method to extract variable names from content, similar to how [_compile_template_string static method.](https://github.com/langfuse/langfuse-python/blob/08298f6fb5b1c33905b5a786a4d1ab0de6cc2ce2/langfuse/model.py#L89) works.

### Usage

```python
text_prompt = TextPromptClient(
    name="example_text_prompt",
    version=1,
    prompt="{{ var1 }} {{ var2 }}",
)

text_variables = text_prompt.variable_names()
print(text_variables)
# > ['var1', 'var2']

chat_prompt = ChatPromptClient(
    name="example_chat_prompt",
    version=1,
    prompt=[
        {"role": "system", "content": "{{ system_var }}."},
        {"role": "user", "content": "{{ user_var }}."},
    ],
)

chat_variables = chat_prompt.variable_names()
print(chat_variables)
# > [{'role': 'system', 'variables': ['system_var']}, {'role': 'user', 'variables': ['user_var']}]
```
ref: https://github.com/orgs/langfuse/discussions/4414

----

> [!IMPORTANT]
> Add `_find_variable_names` method to extract variable names from prompt templates in `BasePromptClient`, with tests for text and chat prompts.
> 
>   - **Behavior**:
>     - Adds `_find_variable_names` static method in `BasePromptClient` to extract variable names from content.
>     - Implements `variable_names()` in `TextPromptClient` and `ChatPromptClient` to return variable names from prompts.
>   - **Tests**:
>     - Adds tests in `test_prompt.py` for `variable_names()` in both text and chat prompts, covering cases with and without variables.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse-python&utm_source=github&utm_medium=referral)<sup> for 6cdbb708d4483d2c1dd6519f2a19931c96375e40. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/langfuse/model.py
+++ b/langfuse/model.py
@@ -1,7 +1,7 @@
 """@private"""
 
 from abc import ABC, abstractmethod
-from typing import Optional, TypedDict, Any, Dict, Union, List
+from typing import Optional, TypedDict, Any, Dict, Union, List, Tuple
 import re
 
 from langfuse.api.resources.commons.types.dataset import (
@@ -59,6 +59,69 @@ class ChatMessageVariables(TypedDict):
     variables: List[str]
 
 
+class TemplateParser:
+    OPENING = "{{"
+    CLOSING = "}}"
+
+    @staticmethod
+    def _parse_next_variable(
+        content: str, start_idx: int
+    ) -> Optional[Tuple[str, int, int]]:
+        """Returns (variable_name, start_pos, end_pos) or None if no variable found"""
+        var_start = content.find(TemplateParser.OPENING, start_idx)
+        if var_start == -1:
+            return None
+
+        var_end = content.find(TemplateParser.CLOSING, var_start)
+        if var_end == -1:
+            return None
+
+        variable_name = content[
+            var_start + len(TemplateParser.OPENING) : var_end
+        ].strip()
+        return (variable_name, var_start, var_end + len(TemplateParser.CLOSING))
+
+    @staticmethod
+    def find_variable_names(content: str) -> List[str]:
+        names = []
+        curr_idx = 0
+
+        while curr_idx < len(content):
+            result = TemplateParser._parse_next_variable(content, curr_idx)
+            if not result:
+                break
+            names.append(result[0])
+            curr_idx = result[2]
+
+        return names
+
+    @staticmethod
+    def compile_template(content: str, data: Dict[str, Any] = {}) -> str:
+        result_list = []
+        curr_idx = 0
+
+        while curr_idx < len(content):
+            result = TemplateParser._parse_next_variable(content, curr_idx)
+
+            if not result:
+                result_list.append(content[curr_idx:])
+                break
+
+            variable_name, var_start, var_end = result
+            result_list.append(content[curr_idx:var_start])
+
+            if variable_name in data:
+                result_list.append(
+                    str(data[variable_name]) if data[variable_name] is not None else ""
+                )
+            else:
+                result_list.append(content[var_start:var_end])
+
+            curr_idx = var_end
+
+        return "".join(result_list)
+
+
 class BasePromptClient(ABC):
     name: str
     version: int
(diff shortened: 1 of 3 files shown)


langfuse/model.py
```diff
@@ -184,6 +219,21 @@ def compile(self, **kwargs) -> List[ChatMessageDict]:
             for chat_message in self.prompt
         ]
 
+    def variable_names(self) -> List[ChatMessageVariables]:
```
hassiebp, line 222: I think the utility of getting a list of variable names of a prompt it to verify before rendering the template whether all vars have been passed. From that lens, I don't understand why having a list of `ChatMessageVariables` has utility for chat prompts rather than a simple list of strings with the variable names. Is there a use case for you here @rhighs ?

langfuse/model.py
```diff
@@ -85,6 +90,28 @@ def get_langchain_prompt(self):
     def _get_langchain_prompt_string(content: str):
         return re.sub(r"{{\s*(\w+)\s*}}", r"{\g<1>}", content)
 
+    @staticmethod
+    def _find_variable_names(content: str) -> List[str]:
```
hassiebp, line 94: Please refactor reused logic into a shared utility. Suggestion by claude (please adapt accordingly)

```python
class TemplateParser:
    OPENING = "{{"
    CLOSING = "}}"

    @staticmethod
    def _parse_next_variable(content: str, start_idx: int) -> Optional[Tuple[str, int, int]]:
        """Returns (variable_name, start_pos, end_pos) or None if no variable found"""
        var_start = content.find(TemplateParser.OPENING, start_idx)
        if var_start == -1:
            return None
            
        var_end = content.find(TemplateParser.CLOSING, var_start)
        if var_end == -1:
            return None
            
        variable_name = content[var_start + len(TemplateParser.OPENING):var_end].strip()
        return (variable_name, var_start, var_end + len(TemplateParser.CLOSING))

    @staticmethod
    def find_variable_names(content: str) -> List[str]:
        names = []
        curr_idx = 0
        
        while curr_idx < len(content):
            result = TemplateParser._parse_next_variable(content, curr_idx)
            if not result:
                break
            names.append(result[0])
            curr_idx = result[2]
            
        return names

    @staticmethod
    def compile_template(content: str, data: Dict[str, Any] = {}) -> str:
        result_list = []
        curr_idx = 0
        
        while curr_idx < len(content):
            result = TemplateParser._parse_next_variable(content, curr_idx)
            
            if not result:
                result_list.append(content[curr_idx:])
                break
                
            variable_name, var_start, var_end = result
            result_list.append(content[curr_idx:var_start])
            
            if variable_name in data:
                result_list.append(str(data[variable_name]) if data[variable_name] is not None else "")
            else:
                result_list.append(content[var_start:var_end])
                
            curr_idx = var_end
            
        return "".join(result_list)
```

hassiebp (changes requested):
Thanks for the PR! Added a two comments before we can get this merged :)

langfuse/model.py
rhighs: Nope, not really. I did it this way to be consistent with the return type you've been using in the compile method above. But at this point I think you're right, a simple list is enough.

rhighs: A reason could be that if someone's using a ChatPromptTemplate he needs some order of organization in his prompts. Variable names in my case are exceptionally useful for post processing purposes, e.g. i know exactly what is the prompt i want to process and then work on it's variables. To me it makes sense having this result organized in this way, otherwise I'd have to manipulate the output in reduction
```python
functools.reduce(operator.iconcat, [
            TemplateParser.find_variable_names(chat_message["content"])
            for chat_message in self.prompt
        ], [])
```

rhighs:
@hassiebp looks good?

hassiebp:
Thanks a lot for your contribution @rhighs - looks good and I'll take it from here! 🙏🏾

PR #1053, opened 2024-12-27
fix: add default value in deprecated argument

Swalloow:
Added a default value of None to the `trace_or_observation` for optional usage.
It prevents potential confusion for users by clearly indicating the optional and deprecated parameter.

----

> [!IMPORTANT]
> Added default value `None` to `trace_or_observation` parameter in `link()` function in `client.py` to indicate it is optional and deprecated.
> 
>   - **Function Signature**:
>     - In `link()` function in `client.py`, added default value `None` to `trace_or_observation` parameter.
>     - This change clarifies that `trace_or_observation` is optional and deprecated.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse-python&utm_source=github&utm_medium=referral)<sup> for 391d2840201adbfe6a9ed9c272a51c65ee33fa91. It will automatically update as commits are pushed.</sup>

### DIFF
--- a/langfuse/client.py
+++ b/langfuse/client.py
@@ -3170,22 +3170,22 @@ class DatasetItemClient:
 
     def link(
         self,
-        trace_or_observation: typing.Union[StatefulClient, str, None],
         run_name: str,
         run_metadata: Optional[Any] = None,
         run_description: Optional[str] = None,
         trace_id: Optional[str] = None,
         observation_id: Optional[str] = None,
+        trace_or_observation: typing.Union[StatefulClient, str, None] = None,
     ):
         """Link the dataset item to observation within a specific dataset run. Creates a dataset run item.
 
         Args:
-            trace_or_observation (Union[StatefulClient, str, None]): The trace or observation object to link. Deprecated: can also be an observation ID.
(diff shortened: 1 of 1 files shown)


marcklingen:
Unfortunately, this was a positional argument in the original api here and later on the keyword arguments were added. @hassiebp can you review?

To me it seems like this requires a major bump as it is a breaking change. If yes, I’d suggest to add this in the next major version as this is clearly tech debt in this interface.

hassiebp (approved):
Thanks for your contribution!

hassiebp:
Apologies, had to revert this [PR here](https://github.com/langfuse/langfuse-python/commit/310a142405b6372fbdd6767b33d70c347f7fdcc2). As Marc has mentioned, the current interface expects positional arguments, such that moving the position of that argument would not be backwards compatible.

hassiebp:
@Swalloow - we have not yet created a v3 branch for which a PR such as this one could be merged into. We have created an internal task though with this PR linked in order to make sure this will be part of our next major release. Thanks again for your contribution!

PR #1087, opened 2025-01-27
check if response has valid json before logging it.

gkmngrgn:
That's a suggestion for -> https://github.com/langfuse/langfuse/issues/4464

Honestly, I'm not sure about if we can consider it as a bug. Because the response comes from outside and there's no guarantee that `res.text` is always json serializable.

There were many options to make the outputs more human-readable but I preferred this way:

- Added a sub function to serialize and log the response at the same time, in each if statement.
- If the status code is 200 or 201, we may not need to the json output, so it logs response text manually, for the specific case.

----

> [!IMPORTANT]
> Improves JSON parsing and error handling in `_process_response` by adding `parse_json_response` function in `langfuse/request.py`.
> 
>   - **Behavior**:
>     - Adds `parse_json_response` function in `_process_response` to handle JSON parsing and logging.
>     - Raises `APIError` for invalid JSON responses in `_process_response`.
>     - Logs response text for status codes 200 and 201 if `return_json` is False.
>   - **Error Handling**:
>     - Improves error handling for non-JSON responses by raising `APIError` with status code and message.
>   - **Logging**:
>     - Logs response text in `_process_response` for successful status codes and when JSON parsing is attempted.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse-python&utm_source=github&utm_medium=referral)<sup> for e2e0e887048446cacd8883723f97b39a43eb880b. It will automatically update as commits are pushed.</sup>


langfuse/request.py
```diff
@@ -82,41 +82,42 @@ def _process_response(
...
-        log.debug("received response: %s", res.text)
+
+        def parse_json_response(response: httpx.Response) -> Dict[str, Any]:
+            try:
+                json_response = response.json()
+            except json.JSONDecodeError:
+                raise APIError(response.status_code, "Invalid JSON response received")
+
+            log.debug("received response: %s", response.text)
+            return json_response
```
hassiebp, line 93: @gkmngrgn thanks for raising this PR! If the parsing fails, we would fall back to None. I would prefer to return `response.text` to have the error reason propagated to the ApiError construction further below

hassiebp (changes requested):
Thanks again for raising this!

langfuse/request.py
gkmngrgn: @hassiebp hey, I did a commit, can you confirm if it's what you meant? https://github.com/langfuse/langfuse-python/pull/1087/commits/36cf9376d4cc6f3bbd96cb67f662bf73d1867848

gkmngrgn: @hassiebp hi again, any update that I can help you in this PR?

PR #1183, opened 2025-05-14
trace is lost in spans coming from async generators

chapati:
## Issue

In the current version of SDK @observe loses context/trace id for spans that are coming from async generators. Such spans appear not on the current trace but added as a new trace and are displayed separately in the UI. The previous implementation have been prematurely removing observations from the stack/not properly restoring the stack state during async generator iteration.

## Impact

My changes improve the reliability of observation tracking when using async generators with the `@observe` decorator. This is particularly important for applications that use async generators extensively, such as streaming responses from language models or other async data processing pipelines.

The fix ensures that observations are properly tracked and that the observation stack is maintained in a consistent state throughout the lifecycle of async generators.


langfuse/decorators/langfuse_decorator.py
```diff
@@ -567,11 +568,21 @@ async def _wrap_async_generator_result(
...
+                    _observation_stack_context.set(stack + [observation])
+
+                    item = await anext(generator)
+                    items.append(item)
+                    _observation_stack_context.set(stack)
+                    yield item
+            except StopAsyncIteration:
+                pass
+            finally:
+                _observation_stack_context.set(stack)
```
mattverantos, line 585: Can you add a test to verify the fix?

langfuse/decorators/langfuse_decorator.py
```diff
@@ -423,6 +423,7 @@ def _finalize_call(
                 observation, result, capture_output, transform_to_string
             )
         elif inspect.isasyncgen(result):
+            #_observation_stack_context.set(stack)
```
mattverantos, line 426: Is this needed?

PR #1222, opened 2025-06-24
feat(prompts): chat message placeholders

nimarb:
> [!IMPORTANT]
> Adds support for chat message placeholders in langfuse-python SDK, enabling dynamic templating in prompts with comprehensive test coverage.
> 
>   - **Behavior**:
>     - Introduces `ChatMessageWithPlaceholders` type in `chat_message_with_placeholders.py` to handle both regular and placeholder messages.
>     - Updates `create_prompt_request.py` and `chat_prompt.py` to use `ChatMessageWithPlaceholders` for prompt creation.
>     - Adds `PlaceholderMessage` model in `placeholder_message.py` for template variables.
>   - **Functionality**:
>     - Modifies `ChatPromptClient` in `model.py` to compile prompts with placeholders, handling unresolved placeholders by logging warnings.
>     - Enhances `get_langchain_prompt()` to convert unresolved placeholders to `MessagesPlaceholder` objects.
>   - **Testing**:
>     - Adds tests in `test_prompt.py` and `test_prompt_compilation.py` to cover placeholder scenarios, including edge cases and integration with Langchain.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse-python&utm_source=github&utm_medium=referral)<sup> for 1c093d29ee370a278c86868650a6269305bcde11. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

----

## Greptile Summary

---
Implements chat message placeholder functionality in the langfuse-python SDK, enabling dynamic templating in chat prompts. This feature is critical for scenarios like injecting conversation history or examples into prompts.

- Added `ChatMessageWithPlaceholders` union type in `langfuse/api/resources/prompts/types/chat_message_with_placeholders.py` supporting both regular chat messages and placeholder messages
- Modified prompt creation to use `ChatMessageWithPlaceholders` instead of `ChatMessage` in `langfuse/api/resources/prompts/types/create_prompt_request.py`
- Added new `PlaceholderMessage` Pydantic model in `langfuse/api/resources/prompts/types/placeholder_message.py` for template variables
- Added comprehensive test coverage in `tests/test_prompt.py` for placeholder scenarios and edge cases


langfuse/model.py
```diff
@@ -276,9 +293,24 @@ def get_langchain_prompt(self, **kwargs) -> str:
 class ChatPromptClient(BasePromptClient):
     def __init__(self, prompt: Prompt_Chat, is_fallback: bool = False):
         super().__init__(prompt, is_fallback)
-        self.prompt = [
-            ChatMessageDict(role=p.role, content=p.content) for p in prompt.prompt
-        ]
+        self.prompt: List[ChatMessageWithPlaceholdersDict] = []
+
+        for p in prompt.prompt:
+            if hasattr(p, "type") and hasattr(p, "name"):
```
hassiebp: we should also add a check for type == placeholder here in case we extend to more types in the future

langfuse/model.py
```diff
@@ -305,15 +339,63 @@ def __eq__(self, other):
...
+
+        Returns:
+            List[ChatMessageDict]: Compiled chat messages
+        """
+        messages_with_placeholders_replaced: List[ChatMessageDict] = []
+
+        # Subsitute the placeholders for their supplied ChatMessages
+        for item in self.prompt:
+            if item["type"] == "placeholder" and item["name"] in placeholders:
+                messages_with_placeholders_replaced.extend(placeholders[item["name"]])
```
hassiebp: another check would be good here to check that `placeholders[item['name']]` is indeed an iterable and raise a meaningful exception if not

langfuse/model.py
```diff
@@ -289,6 +321,7 @@ def compile(self, **kwargs) -> List[ChatMessageDict]:
                 role=chat_message["role"],
             )
             for chat_message in self.prompt
+            if chat_message["type"] == "message"
```
hassiebp: we are implicitly filtering out the placeholders here - we probably should keep them in there. What do you think? User will get then an explicitly error by the LLM and will know to use compileWithPlaceholders.

Alternatively / additionally we can log a warning that compile was used even though placeholders were present

langfuse/model.py
```diff
@@ -305,15 +339,63 @@ def __eq__(self, other):
...
                     for m1, m2 in zip(self.prompt, other.prompt)
                 )
                 and self.config == other.config
             )
 
         return False
 
+    def compileWithPlaceholders(
+        self,
+        variables: Dict[str, Any],
```
hassiebp: the value must be a string here too Dict[str, str]

langfuse/model.py
```diff
@@ -305,15 +339,63 @@ def __eq__(self, other):
...
+        return [
+            ChatMessageDict(
+                content=TemplateParser.compile_template(
+                    chat_message["content"],
+                    variables,
+                ),
+                role=chat_message["role"],
+            )
+            for chat_message in messages_with_placeholders_replaced
+            if "role" in chat_message and "content" in chat_message
```
hassiebp: we would filter out passed messages here that are not having 'role' and 'content' key. We should keep them, as the user has explicitly passed them into the placeholder

langfuse/model.py
```diff
@@ -305,15 +339,63 @@ def __eq__(self, other):
...
+        ]
+
     def get_langchain_prompt(self, **kwargs):
```
hassiebp: we need to cover langchain prompts as well where users are not compiling prompts with langfuse but with langchain. In this case, they might want to only fill in placeholders and keep variables untouched and then call get_langchain_prompt.

We can either add a separate method that is only handling replacing placeholders, or we rely on `compileWithPlaceholders(variables={}, placeholders=...)` but we need a test here to make sure this works as expected

langfuse/model.py
```diff
@@ -276,40 +302,175 @@ def get_langchain_prompt(self, **kwargs) -> str:
...
+                    chat_message["content"],
+                    kwargs,
                 ),
                 role=chat_message["role"],
             )
             for chat_message in self.prompt
+            if "content" in chat_message and "role" in chat_message
         ]
 
+    def set(self, placeholders: Dict[str, List[ChatMessageDict]]) -> "ChatPromptClient":
```
hassiebp: What do we need this method for?

langfuse/model.py
```diff
@@ -276,40 +302,161 @@ def get_langchain_prompt(self, **kwargs) -> str:
...
+
+        Args:
+            placeholders: Dictionary mapping placeholder names to lists of chat messages
+
+        Returns:
+            ChatPromptClient: Self for method chaining
+        """
+        self.placeholder_fillins = placeholders.copy()
+        return self
+
```
hassiebp, line 391: keyword argument

PR #1233, opened 2025-06-30
Fix async generator case in decorator

kouk:
The decorator code in `langfuse/_client/observe.py` seems to have been built to support the case of observing an async generator however it misses the mark by using `asyncio.iscoroutinefunction` to decide whether to wrap it in the async or the sync implementation. That method doesn't detect async generators.

The solution is to use the inspect module and check for both cases.

### DIFF
--- a/langfuse/_client/observe.py
+++ b/langfuse/_client/observe.py
@@ -1,4 +1,3 @@
-import asyncio
 import inspect
 import logging
 import os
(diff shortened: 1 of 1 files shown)


kouk:
related to https://github.com/langfuse/langfuse/issues/7226

hassiebp:
Released in Python SDK v3.2.1

PR #1249, opened 2025-07-09
feat(client): add optional override default cache ttl capability for prompt caching

cemasma:
**Summary**
This PR enhances the Langfuse Python SDK’s prompt caching mechanism by enabling runtime overrides of the default 60-second TTL (default_cache_ttl_seconds). It allows developers to optionally specify a custom TTL—or disable caching entirely—during the initialization of the Langfuse client.

**Goal**
- To allow users to control the default prompt cache config without specifying it for every get prompt request.
- You can extend the cache to reduce API calls (e.g., set to 300 seconds) or you can disable caching by setting it to 0 for immediate prompt fetches every time at initial.


langfuse/_client/client.py
```diff
@@ -108,6 +108,7 @@ class Langfuse:
...
         additional_headers (Optional[Dict[str, str]]): Additional headers to include in all API requests and OTLPSpanExporter requests. These headers will be merged with default headers. Note: If httpx_client is provided, additional_headers must be set directly on your custom httpx_client as well.
         tracer_provider(Optional[TracerProvider]): OpenTelemetry TracerProvider to use for Langfuse. This can be useful to set to have disconnected tracing between Langfuse and other OpenTelemetry-span emitting libraries. Note: To track active spans, the context is still shared between TracerProviders. This may lead to broken trace trees.
+        default_cache_ttl_seconds (Optional[int]): Override the default cache TTL (time-to-live) of 60 seconds with a custom value (in seconds) for prompts.
```
hassiebp: Can we rename this to `prompt_cache_default_ttl_seconds`? This makes it clearer

hassiebp (changes requested):
Thanks for your contribution @cemasma !

Do you really need to configure different default prompt cache durations during runtime? If not, we could opt for a simpler approach where we allow a env var `LANGFUSE_PROMPT_CACHE_DEFAULT_TTL_SECONDS` and simply check in the prompt cache init whether it is set or passed

cemasma:
> Thanks for your contribution @cemasma !
> 
> Do you really need to configure different default prompt cache durations during runtime? If not, we could opt for a simpler approach where we allow a env var `LANGFUSE_PROMPT_CACHE_DEFAULT_TTL_SECONDS` and simply check in the prompt cache init whether it is set or passed

thanks for the review @hassiebp just updated it to fetch value from the environment and definitely it turned out much simpler now

hassiebp:
Thanks for your contribution, @cemasma !

PR #1267, opened 2025-07-21
fix: add support for capturing usage details with ChatBedRock

pbvignesh:
This PR is a fix for [Issue 7657](https://github.com/langfuse/langfuse/issues/7657) where the usage details for ChatbedRock were not captured properly while it was working fine for ChatbedRockConverse. 

I have mentioned the cause for the bug in the issue comment but I am pasting the gist of it here. 

### Cause of the Bug

The error for this seems to be at **langchain/CallbackHandler.py** due to mainly two reasons

- There is no key mapping for converting between the model key to the langfuse key
- The model keys for ChatBedrock is the exact same as the OpenAI keys which we are skipping since they are being handled server side (I do not know exactly where it is being handled server side and would love to get your help here)

The problem specifically is with the function **__parse_usage_model_**.

When it comes to ChatBedrockConverse, the input passed to this function looks like the following 

```
{'input_tokens': 11, 'output_tokens': 30, 'total_tokens': 41, 'input_token_details': {'cache_creation': 0, 'cache_read': 0}}
```

while with ChatBedrock, the input passed to this function looks like the following 

```
{'prompt_tokens': 11, 'completion_tokens': 30, 'cache_read_input_tokens': 0, 'cache_write_input_tokens': 0, 'total_tokens': 41}
```

### Fix for the Bug

- I've added the key mapping for converting between the model key to the langfuse key
- Only the input, output and total keys for ChatBedrock is the same as OpenAI so I've added the remaining keys in OpenAI for the code to be able to differentiate between OpenAI vs other providers 

Here is a screenshot of the issue before fixing this

issue before fix

Here is a screenshot of the trace after fixing this issue

issue post fix

I tried running the existing tests but some of them are failing because they require me to provide an OpenAI API key but I don't think they have a bearing on this fix. But do let me know if the team requires me to do some additional testing beyond this.

----

> [!IMPORTANT]
> Fixes usage details capture for ChatBedRock by adding key mappings and differentiating keys from OpenAI in `CallbackHandler.py`.
> 
>   - **Behavior**:
>     - Fixes usage details capture for ChatBedRock in `_parse_usage_model()` in `CallbackHandler.py`.
>     - Adds key mappings for `prompt_tokens` to `input` and `completion_tokens` to `output`.
>     - Differentiates ChatBedRock keys from OpenAI by adding `completion_token_details` and `prompt_token_details` to OpenAI keys.
>   - **Misc**:
>     - Addresses [Issue 7657](https://github.com/langfuse/langfuse/issues/7657).
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse-python&utm_source=github&utm_medium=referral)<sup> for 09e0acf646e3b2fd32208a3a88b96aa3157a2766. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

---

## Greptile Summary

This PR addresses a bug in the LangChain callback handler where token usage metrics were not being properly captured for Amazon's ChatBedrock model. The issue arose because ChatBedrock uses similar token usage keys to OpenAI (`prompt_tokens`, `completion_tokens`), which were being skipped as they were assumed to be OpenAI-specific formats handled server-side.

The fix adds proper key mapping for ChatBedrock's format and improves the detection of genuine OpenAI responses by checking for additional OpenAI-specific keys (`completion_token_details`, `prompt_token_details`). This ensures ChatBedrock's usage metrics are properly captured while maintaining special handling for OpenAI responses.

## Confidence score: 4/5

1. The changes look safe as they only impact token usage reporting for ChatBedrock and preserve existing behavior
2. The fix is well-reasoned and addresses the root cause by distinguishing between OpenAI and ChatBedrock formats
3. Files needing attention:
   - `langfuse/langchain/CallbackHandler.py`: The changes look good but should be tested with both ChatBedrock and OpenAI to ensure no regressions

### DIFF
--- a/langfuse/langchain/CallbackHandler.py
+++ b/langfuse/langchain/CallbackHandler.py
@@ -1,16 +1,18 @@
 import typing
+from contextvars import Token
 
 import pydantic
+from opentelemetry import context, trace
 
-from langfuse._client.get_client import get_client
 from langfuse._client.attributes import LangfuseOtelSpanAttributes
+from langfuse._client.get_client import get_client
 from langfuse._client.span import (
-    LangfuseGeneration,
-    LangfuseSpan,
     LangfuseAgent,
     LangfuseChain,
-    LangfuseTool,
+    LangfuseGeneration,
     LangfuseRetriever,
+    LangfuseSpan,
+    LangfuseTool,
 )
 from langfuse.logger import langfuse_logger
 
@@ -86,6 +88,7 @@ class LangchainCallbackHandler(LangchainBaseCallbackHandler):
                 LangfuseRetriever,
             ],
         ] = {}
+        self.context_tokens: Dict[UUID, Token] = {}
         self.prompt_to_parent_run_map: Dict[UUID, Any] = {}
         self.updated_completion_start_time_memo: Set[UUID] = set()
 
@@ -210,11 +213,14 @@ class LangchainCallbackHandler(LangchainBaseCallbackHandler):
             if run_id is None or run_id not in self.runs:
                 raise Exception("run not found")
 
-            self.runs[run_id].update(
-                level="ERROR",
-                status_message=str(error),
-                input=kwargs.get("inputs"),
-            ).end()
+            observation = self._detach_observation(run_id)
+
+            if observation is not None:
+                observation.update(
+                    level="ERROR",
+                    status_message=str(error),
+                    input=kwargs.get("inputs"),
+                ).end()
 
         except Exception as e:
             langfuse_logger.exception(e)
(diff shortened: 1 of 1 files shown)


pbvignesh:
Hi @hassiebp just checking in to see if this is just put on hold due to bandwidth issues from the langfuse team or if something is missing. If you need me to look into additional stuff or do other research for this issue or any other issue do let me know, I'm interested in helping out

langfuse/langchain/CallbackHandler.py
```diff
@@ -865,10 +868,16 @@ def _parse_usage_model(usage: typing.Union[pydantic.BaseModel, dict]) -> Any:
...
-        openai_key in usage_model
-        for openai_key in ["prompt_tokens", "completion_tokens", "total_tokens"]
-    ):
+    openai_keys = [
+        "prompt_tokens",
+        "completion_tokens",
+        "total_tokens",
+        # We add the below two since OpenAI keys look similar to ChatBedRock
+        # keys except for the below two keys
+        "completion_token_details",
```
hassiebp: For OpenAI this should be `completion_tokens_detail` and `prompt_tokens_details`: https://platform.openai.com/docs/api-reference/chat/create

pbvignesh: Thanks ! My code was a bit outdated apparently and I have merged main with this which has given the updated keys anyway !

hassiebp:
Thanks for sharing! In the meantime a similar fix has already hit the main branch. Could you please upgrade and try again whether you are running into issues?

pbvignesh:
Hey @hassiebp I just did a quick check now in my system and I can confirm that the issue still persists. I've tested this with langfuse python sdk version **3.3.1** and the tokens still don't get tracked
Screenshot 2025-08-26 at 3 34 22 PM

This is because the mapping between the ChatBedRock and langfuses's input and output tokens are still not mapped to each other in the conversion_list variable in the __parse_usage_model_ function in the latest version I think which my fix addresses.

Let me know if this makes sense.

PR #1268, opened 2025-07-22
chore: add claude file

nimarb:


.claude/settings.json
```diff
@@ -0,0 +1,14 @@
+{
+  "permissions": {
+    "allow": [
```
hassiebp, line 3: Thanks for adding these! Can we remove the permissions from the repo?

I prefer to keep the permissions up to each developer rather than having project wide defaults. For some commands, a developer confirmation should be done. E.g. here we are allowing all `cat` commands for all files, claude might decide to read .env (`cat .env`) and send to LLM which we can catch if it requires dev confirmation

hassiebp: You can track your local permissions in a settings.local.json

PR #1275, opened 2025-07-29
fix: avoid private import of Langfuse to satisify Pyright

janwinkler1:
Fixes https://github.com/langfuse/langfuse/issues/8176

Pyright flags Langfuse import as private after py.typed was added. 

Fix: explicitly re-export in __init__.py


hassiebp:
Thanks for your contribution @janwinkler1 !

PR #1297, opened 2025-08-14
fix(client): skip string serialization in attribute values

hassiebp:


langfuse/_client/attributes.py
```diff
@@ -152,7 +152,10 @@ def create_generation_attributes(
 
 
 def _serialize(obj: Any) -> Optional[str]:
-    return json.dumps(obj, cls=EventSerializer) if obj is not None else None
+    if obj is None or isinstance(obj, str):
+        return obj
+
+    return json.dumps(obj, cls=EventSerializer)
```
j3r0lin, line 158: ensure_ascii here should be False

PR #1312, opened 2025-08-27
fix(langchain): do not end span on agent callbacks

hassiebp:


langfuse/langchain/CallbackHandler.py
```diff
@@ -401,19 +397,17 @@ def on_agent_action(
...
+            if agent_run is not None:
                 agent_run._otel_span.set_attribute(
                     LangfuseOtelSpanAttributes.OBSERVATION_TYPE, "agent"
                 )
 
-            agent_run.update(
-                output=action,
-                input=kwargs.get("inputs"),
-            ).end()
+                agent_run.update(
```
qnnn, line 407: Hi @hassiebp , I have a question. Should we update the input and output during the agent action or finish process? It’s an intermediate step and may not accurately represent the true input and output of the span.

hassiebp: Could you please try with the latest SDK version and report whether you see unexpected results?

qnnn: > Could you please try with the latest SDK version and report whether you see unexpected results?

This appears to be a different issue that I didn't explain clearly before (or it might not be an issue at all).

+ In the agent action callback when setting output, if the tool execution fails, the current output is the action's output, but I expect the output to be empty in such cases.

PR #1314, opened 2025-08-28
Add tool_calls in AIMessage

nicolasse:


nicolasse:
Screenshot 2025-08-28 at 7 35 04 PM

Based on the langfuse JS package, the Python version is missing tool_calls in AIMessage.
This causes the playground to break when the agent makes a tool call.
Screenshot 2025-08-28 at 7 37 33 PM

nicolasse:
Screenshot 2025-08-28 at 7 51 37 PM
This image is after the fix, we can see the Assistant message is different

langfuse/langchain/CallbackHandler.py
```diff
@@ -912,6 +912,13 @@ def _convert_message_to_dict(self, message: BaseMessage) -> Dict[str, Any]:
             message_dict = {"role": "user", "content": message.content}
         elif isinstance(message, AIMessage):
             message_dict = {"role": "assistant", "content": message.content}
+
+            if (
+                hasattr(message, "tool_calls")
+                and message.tool_calls
```
hassiebp:
```suggestion
                and message.tool_calls is not None
```

PR #1334, opened 2025-09-11
feat(experiments): add experiment runner

hassiebp:
> [!IMPORTANT]
> Adds an experiment runner to the Langfuse Python SDK for evaluating AI tasks with automatic tracing and scoring, including async execution support and extensive test coverage.
> 
>   - **Behavior**:
>     - Adds `run_experiment` method to `client.py` for running experiments with automatic tracing and evaluation.
>     - Supports concurrent task execution with semaphore-based concurrency control.
>     - Handles errors gracefully, isolating individual failures.
>   - **Modules**:
>     - New `experiments.py` module with type definitions for `TaskFunction`, `EvaluatorFunction`, and `RunEvaluatorFunction`.
>     - `types.py` updated to expose experiment types as a public API.
>     - `datasets.py` updated to integrate dataset objects with experiments.
>   - **Utilities**:
>     - Adds `run_async_safely` function in `utils.py` to handle async execution safely.
>   - **Tests**:
>     - Extensive tests added in `test_experiments.py` to validate experiment functionality.
>     - Tests for utility functions in `test_utils.py` to ensure async handling works correctly.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse-python&utm_source=github&utm_medium=referral)<sup> for 9e7cac693f8d900c8b0c68bda46a1bac9fce6ac7. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

----

---

## Greptile Summary

Updated On: 2025-09-11 16:02:25 UTC

This review covers only the changes made since the last review, not the entire PR. The previous review identified two critical issues: a missing `statistics` import and a type inconsistency with an evaluator returning a plain dict instead of an Evaluation object. These issues remain unaddressed in the current version.

The experiment runner framework introduces substantial new functionality to the Langfuse Python SDK, enabling systematic evaluation of AI tasks on datasets with automatic tracing and scoring. The core implementation centers around a new `run_experiment` method in the main Langfuse client that handles concurrent execution of tasks and evaluators while creating proper traces and persisting results.

The new `langfuse/_client/experiments.py` module establishes the foundational type system with Protocol definitions for `TaskFunction`, `EvaluatorFunction`, and `RunEvaluatorFunction`, along with comprehensive TypedDict structures for data handling. The `langfuse/types.py` file transforms from a private module to a public API, exposing experiment types through a controlled export interface. The dataset integration in `langfuse/_client/datasets.py` provides a convenience method for running experiments directly on dataset objects.

The implementation includes semaphore-based concurrency control, comprehensive error handling that isolates individual failures, and extensive test coverage validating both local and Langfuse-hosted dataset scenarios. The async architecture efficiently handles large datasets while respecting API rate limits, and automatic score persistence integrates seamlessly with the Langfuse platform.

## Confidence score: 3/5

- This PR has significant implementation issues that prevent safe merging, including missing imports that will cause runtime errors
- Score lowered due to unresolved critical issues from previous review: missing statistics import and type inconsistency in evaluator returns
- Pay close attention to the import statements in `client.py` and `experiments.py`, and the evaluator return types in `test_experiments.py`

### Context used:
**Rule -** Open a GitHub issue or discussion first before submitting PRs to explain the rationale and necessity of the proposed changes, as required by the contributing guide. ([link](https://app.greptile.com/review/custom-context?memory=f66422d7-e71a-47d4-928b-df849c888300))

### DIFF
--- a/langfuse/_client/client.py
+++ b/langfuse/_client/client.py
@@ -101,6 +101,7 @@ from langfuse.model import (
     ChatPromptClient,
     CreateDatasetItemRequest,
     CreateDatasetRequest,
+    CreateDatasetRunItemRequest,
     Dataset,
     DatasetItem,
     DatasetStatus,
@@ -2806,8 +2807,6 @@ class Langfuse:
                 # Link to dataset run if this is a dataset item
                 if hasattr(item, "id") and hasattr(item, "dataset_id"):
                     try:
-                        from langfuse.model import CreateDatasetRunItemRequest
-
                         dataset_run_item = self.api.dataset_run_items.create(
                             request=CreateDatasetRunItemRequest(
                                 runName=experiment_run_name,
@@ -2815,6 +2814,7 @@ class Langfuse:
                                 metadata=experiment_metadata,
                                 datasetItemId=item.id,  # type: ignore
                                 traceId=trace_id,
+                                observationId=span.id,
                             )
                         )
 


langfuse/_client/client.py
```diff
@@ -2444,6 +2458,411 @@ def get_dataset(
...
+            evaluators: List of functions to evaluate each item's output individually.
+                Each evaluator receives input, output, expected_output, and metadata.
+                Can return single Evaluation dict or list of Evaluation dicts.
+            run_evaluators: List of functions to evaluate the entire experiment run.
+                Each run evaluator receives all item_results and can compute aggregate metrics.
+                Useful for calculating averages, distributions, or cross-item comparisons.
+            max_concurrency: Maximum number of concurrent task executions (default: 50).
+                Controls the number of items processed simultaneously. Adjust based on
+                API rate limits and system resources.
+            metadata: Optional metadata dictionary to attach to all experiment traces.
```
marcklingen, line 2513: there is also datasetrun metadata that we exposed via "link" and then upserted on every item, we moved to trace metadata here to make this available on non-dataset experiments, correct?

langfuse/_client/client.py
```diff
@@ -2444,6 +2458,411 @@ def get_dataset(
...
+                This metadata will be included in every trace created during the experiment.
+
+        Returns:
+            ExperimentResult dictionary containing:
```
marcklingen: does it make sense to make this a class? could then have the pretty print method.

no strong take here, could just make adding logic and discovering what to do with the result easier

I'd consider switching from dict to class breaking here if we want to change it later

langfuse/_client/client.py
```diff
@@ -2444,6 +2458,411 @@ def get_dataset(
...
+                # Link to dataset run if this is a dataset item
+                if hasattr(item, "id") and hasattr(item, "dataset_id"):
+                    try:
+                        from langfuse.model import CreateDatasetRunItemRequest
+
+                        dataset_run_item = self.api.dataset_run_items.create(
+                            request=CreateDatasetRunItemRequest(
+                                runName=experiment_name,
+                                runDescription=experiment_description,
+                                metadata=experiment_metadata,
```
marcklingen, line 2814: doc of method does not mention that the metadata is also added here, add to docstring?

langfuse/_client/client.py
```diff
@@ -2444,6 +2458,411 @@ def get_dataset(
...
+                                datasetItemId=item.id,  # type: ignore
+                                traceId=trace_id,
```
marcklingen, line 2816: for performance reasons could also make sense to also link to observation here, @maxdeichmann and @Steffen911? (chatted with them about this today given potential backend changes)

langfuse/_client/client.py
hassiebp: Yes, correct! We spread out all the experiment metadata on every resulting trace on the first level (no longer the run_metadata key) to allow users to filter efficiently for it in UI

langfuse/_client/client.py
hassiebp: Agree - will move this to a class now as converting this later will be fiddly (though possible)

langfuse/_client/client.py
Steffen911: If there is a specific span_id we know about that would be amazing!

hassiebp: I think what would be useful here eventually is that **all** resulting spans will hold an attribute that tracks to what dataset run they belong similar as we are planning for session_id, user_id etc. That should be part of that change and not yet in this PR. 

Would you agree @maxdeichmann ?

marcklingen: > I think what would be useful here eventually is that **all** resulting spans will hold an attribute that tracks to what dataset run they belong similar as we are planning for session_id, user_id etc. That should be part of that change and not yet in this PR.<br>

Agree that this could make sense. We could here already include the span id (we will need to support trace+observation id anyway and it will be more performant that only trace id)

hassiebp: Added ✅

maxdeichmann: @hassiebp i agree we should track both and we should also pass the experiment identifiers onto all spans within a trace. I do not see here that you add the observationid, right?

langfuse/_client/client.py
```diff
@@ -2444,6 +2459,411 @@ def get_dataset(
...
+                            input=input_data,
+                            output=output,
+                            expected_output=expected_output,
+                            metadata=eval_metadata,
+                        )
+                        evaluations.extend(eval_results)
+
+                        # Store evaluations as scores
+                        for evaluation in eval_results:
+                            self.create_score(
```
bdepebhe, line 2856: The `data_type` attribute of `Evaluation` is unused for item-level evaluators, despite being showcased in the docstring of `Evaluation` with an example of item-level CATEGORICAL evaluator. This is misleading for users.
For example, as I tested with BOOLEAN evaluator, I expected True or False, and got 1 and -1

langfuse/_client/client.py
```diff
@@ -2444,6 +2459,411 @@ def get_dataset(
...
+                                trace_id=trace_id,
+                                name=evaluation.name,
+                                value=evaluation.value or -1,
```
bdepebhe, line 2859: This conversion to -1 is problematic for numerical scores. `0.0` can be a legitimate value for a score (example: 0. as a Pearson correlation coefficient means no correlation, but -1 means a perfect negative correlation !)

PR #1336, opened 2025-09-15
fix(v2-stable): task_done exception on dropped item

newmanifold:
**Problem:**
Sometimes for large traces we are seeing exception `ValueError('task_done() called too many times')`, After few of these exceptions , traces/spans aren't getting into langfuse.

**Cause:**
In file [ingestion_consumer](https://github.com/langfuse/langfuse-python/blob/v2-stable/langfuse/_task_manager/ingestion_consumer.py) dropped events inside _truncate_item_in_place were being appended to the batch events in the _next method, which caused the task_done call on the ingestion queue in the upload method to fail with an exception.
The issue is that task_done is already called for dropped items inside _truncate_item_in_place.

**Changes:**
- Added missing metadata check to event drop condition and updated warning.

----

> [!IMPORTANT]
> Fixes `task_done` exception in `ingestion_consumer.py` by preventing dropped items from being appended to batch events.
> 
>   - **Behavior**:
>     - Fixes `ValueError('task_done() called too many times')` in `ingestion_consumer.py` by preventing dropped items from being appended to batch events.
>     - Adds a flag in `_truncate_item_in_place()` to indicate if an item was dropped.
>     - Modifies `_next()` to append only non-dropped items to events.
>   - **Functions**:
>     - `_truncate_item_in_place()` now returns a tuple `(item_size, dropped)`.
>     - Updates `_next()` to handle the new return value from `_truncate_item_in_place()` and conditionally append events.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse-python&utm_source=github&utm_medium=referral)<sup> for bffe753272254e380ad963bdef0903ee58970db7. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

---

## Greptile Summary

Updated On: 2025-09-15 05:00:00 UTC

This PR fixes a critical bug in the task queue management system for Langfuse's ingestion consumer. The issue occurred when processing large traces that exceeded size limits - the `_truncate_item_in_place()` method would drop oversized events and call `task_done()` on the ingestion queue, but these dropped events were still being added to the batch for processing. Later, when the `upload()` method processed the batch, it would call `task_done()` again for each event in the batch, including the already-dropped ones, resulting in `ValueError('task_done() called too many times')`.

The fix modifies the `_truncate_item_in_place()` method to return a tuple containing both the item size and a boolean flag indicating whether the item was dropped. The `_next()` method now checks this flag and only appends non-dropped events to the batch. This ensures that `task_done()` is called exactly once per queue item - either during the truncation/drop process or during batch processing, but never both.

The change maintains backward compatibility in terms of functionality while fixing the queue state management. The ingestion consumer continues to handle oversized events by truncating or dropping them as before, but now properly tracks which items were dropped to prevent double-counting in the task queue. This fix is essential for maintaining the integrity of Langfuse's async task processing system, particularly when dealing with large traces that require size-based filtering.

## Confidence score: 4/5

- This PR addresses a well-defined concurrency bug with a focused solution that maintains existing behavior
- Score reflects solid understanding of the queue management issue and appropriate fix, though the async nature of task queues can have subtle edge cases
- Pay close attention to the tuple return modification in `_truncate_item_in_place()` and its usage in `_next()`

### DIFF
--- a/langfuse/_task_manager/ingestion_consumer.py
+++ b/langfuse/_task_manager/ingestion_consumer.py
@@ -122,7 +122,7 @@ class IngestionConsumer(threading.Thread):
                 self._media_manager.process_media_in_event(event)
 
                 # truncate item if it exceeds size limit
-                item_size, dropped = self._truncate_item_in_place(
+                item_size = self._truncate_item_in_place(
                     event=event,
                     max_size=MAX_EVENT_SIZE_BYTES,
                     log_message="<truncated due to size exceeding limit>",
@@ -137,8 +137,7 @@ class IngestionConsumer(threading.Thread):
 
                     continue
 
-                if not dropped:
-                    events.append(event)
+                events.append(event)
 
                 total_size += item_size
                 if total_size >= MAX_BATCH_SIZE_BYTES:
@@ -167,11 +166,8 @@ class IngestionConsumer(threading.Thread):
         event: Any,
         max_size: int,
         log_message: Optional[str] = None,
-    ) -> tuple[int, bool]:
-        """Truncate the item in place to fit within the size limit.
-
-        Returns a tuple of the item size and a boolean indicating if the item was dropped.
-        """
+    ) -> int:
+        """Truncate the item in place to fit within the size limit."""
         item_size = self._get_item_size(event)
         self._log.debug(f"item size {item_size}")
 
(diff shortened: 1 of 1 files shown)


hassiebp:
In that case, this PR could be simplified to the drop condition to be extended to check for missing metadata as well: so event is only dropped if input, output, and metadata are missing

newmanifold:
Hey! 
Do you mean we should only add a check for missing metadata in the drop condition and remove all the other changes in the PR?
That would solve the issue in my specific case, but if the drop condition can still be triggered, then we’d run into the same problem of task_done being called multiple times.

Unless, of course, there's no actual way for the drop condition to be triggered in which case, the drop condition itself might be redundant.

newmanifold:
@hassiebp I've done the changes, let me know if its fine or anything else is needed to be done.
I've also updated the PR description for the changes but Ellipses and Greptile Summaries are still older, I hope that's okay.

PR #1346, opened 2025-09-17
fix(prompt-placeholders): don't strip tool calls from placeholder messages

nimarb:
> [!IMPORTANT]
> Fixes bug in `ChatPromptClient.compile()` to preserve tool calls in message placeholders, with a new test added for verification.
> 
>   - **Behavior**:
>     - Fixes bug in `ChatPromptClient.compile()` to preserve tool calls in message placeholders.
>     - Ensures all fields from original messages are retained, including `role` and `content`.
>   - **Tests**:
>     - Adds `test_tool_calls_preservation_in_message_placeholder()` in `test_prompt_compilation.py` to verify tool calls are preserved during compilation.
> 
> <sup>This description was created by </sup>[Ellipsis](https://www.ellipsis.dev?ref=langfuse%2Flangfuse-python&utm_source=github&utm_medium=referral)<sup> for ab40d06c6e3f80f16f69cb4ce4ee4f976a324664. You can [customize](https://app.ellipsis.dev/langfuse/settings/summaries) this summary. It will automatically update as commits are pushed.</sup>

---

## Greptile Summary

Updated On: 2025-09-17 12:24:14 UTC

This PR fixes a bug in the Langfuse prompt system where message placeholders were losing important OpenAI message fields like `tool_calls` during compilation. The change affects the prompt compilation logic in `langfuse/model.py` and adds a corresponding test in `tests/test_prompt_compilation.py`.

**The Core Issue**: The previous implementation of placeholder message compilation only preserved `role` and `content` fields by creating a restrictive `ChatMessageDict` structure. This caused data loss for complex chat messages that contain additional metadata like `tool_calls`, `function_call`, or other OpenAI-specific properties.

**The Solution**: The fix modifies the compilation logic to preserve all fields from the original message dictionary while ensuring backward compatibility. Instead of creating a new `ChatMessageDict` with only role and content, the code now:
1. Creates a copy of the entire original message dictionary 
2. Ensures `role` and `content` fields are always present with sensible defaults
3. Compiles template variables within the content field
4. Preserves all other fields unchanged

**Integration with Codebase**: This change fits into Langfuse's prompt management system, which supports message placeholders for conversation history injection. The `BasePromptClient` and `ChatPromptClient` classes are part of the framework's template compilation pipeline, where prompts can contain variable placeholders (`{{variable_name}}`) that get resolved at runtime. The fix ensures that when message objects are used as placeholder values, their complete structure is maintained.

**Type System Updates**: The return type annotations for the `compile` method in both abstract and concrete classes were updated to include `Dict[str, Any]` alongside existing types, reflecting that compiled messages can now contain arbitrary fields beyond the basic `ChatMessageDict` structure.

## Confidence score: 4/5

- This PR addresses a clear bug with a targeted fix that preserves backward compatibility
- The solution follows the simpler approach principle by extending existing logic rather than adding complex state tracking
- Comprehensive test coverage ensures the fix works correctly and prevents regression
- Pay close attention to the type annotation changes and ensure downstream consumers can handle the expanded return types

### Context used:
**Rule -** Prefer simpler solutions when fixing bugs - if the root cause can be addressed by extending an existing condition rather than adding complex state tracking, choose the simpler approach. ([link](https://app.greptile.com/review/custom-context?memory=165049f5-c158-4d70-9930-55fb9b833eb9))

### DIFF
--- a/langfuse/__init__.py
+++ b/langfuse/__init__.py
@@ -1,21 +1,23 @@
 """.. include:: ../README.md"""
 
+from langfuse.experiment import Evaluation
+
 from ._client import client as _client_module
 from ._client.attributes import LangfuseOtelSpanAttributes
 from ._client.constants import ObservationTypeLiteral
 from ._client.get_client import get_client
 from ._client.observe import observe
 from ._client.span import (
-    LangfuseEvent,
-    LangfuseGeneration,
-    LangfuseSpan,
     LangfuseAgent,
-    LangfuseTool,
     LangfuseChain,
     LangfuseEmbedding,
     LangfuseEvaluator,
-    LangfuseRetriever,
+    LangfuseEvent,
+    LangfuseGeneration,
     LangfuseGuardrail,
+    LangfuseRetriever,
+    LangfuseSpan,
+    LangfuseTool,
 )
 
 Langfuse = _client_module.Langfuse
@@ -36,4 +38,7 @@ __all__ = [
     "LangfuseEvaluator",
     "LangfuseRetriever",
     "LangfuseGuardrail",
+    "Evaluation",
+    "experiment",
+    "api",
 ]
--- a/langfuse/_client/client.py
+++ b/langfuse/_client/client.py
@@ -3,6 +3,7 @@
 This module implements Langfuse's core observability functionality on top of the OpenTelemetry (OTel) standard.
 """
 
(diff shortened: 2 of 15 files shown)


nimarb:
should fix: https://github.com/langfuse/langfuse/issues/9051

PR #1385, opened 2025-09-29
feat: propagate trace attributes onto all child spans on update

Steffen911:
Replaces https://github.com/langfuse/langfuse-python/pull/1349

### DIFF
--- a/langfuse/_client/client.py
+++ b/langfuse/_client/client.py
@@ -4,6 +4,7 @@ This module implements Langfuse's core observability functionality on top of the
 """
 
 import asyncio
+import json
 import logging
 import os
 import re
@@ -27,8 +28,11 @@ from typing import (
 
 import backoff
 import httpx
-from opentelemetry import trace
-from opentelemetry import trace as otel_trace_api
+from opentelemetry import (
+    trace as otel_trace_api,
+    baggage as otel_baggage_api,
+    context as otel_context_api,
+)
 from opentelemetry.sdk.trace import TracerProvider
 from opentelemetry.sdk.trace.id_generator import RandomIdGenerator
 from opentelemetry.util._decorator import (
@@ -110,6 +114,11 @@ from langfuse.model import (
     TextPromptClient,
 )
 from langfuse.types import MaskFunction, ScoreDataType, SpanLevel, TraceContext
+
+# Context key constants for Langfuse context propagation
+LANGFUSE_CTX_USER_ID = "langfuse.ctx.user.id"
+LANGFUSE_CTX_SESSION_ID = "langfuse.ctx.session.id"
+LANGFUSE_CTX_METADATA = "langfuse.ctx.metadata"
 
 
 class Langfuse:
@@ -1667,6 +1676,11 @@ class Langfuse:
                 span.update(output=response)
             ```
         """
+        warnings.warn(
+            "update_current_trace is deprecated and will be removed in a future version. ",
+            DeprecationWarning,
+            stacklevel=2,
+        )
         if not self._tracing_enabled:
             langfuse_logger.debug(
                 "Operation skipped: update_current_trace - Tracing is disabled or client is in no-op mode."
@@ -1811,7 +1825,7 @@ class Langfuse:
             is_remote=False,
         )
 
-        return trace.NonRecordingSpan(span_context)
+        return otel_trace_api.NonRecordingSpan(span_context)
 
     def _is_valid_trace_id(self, trace_id: str) -> bool:
         pattern = r"^[0-9a-f]{32}$"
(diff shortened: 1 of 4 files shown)


langfuse/_client/client.py
```diff
@@ -3450,3 +3467,211 @@ def clear_prompt_cache(self) -> None:
...
+        if as_baggage:
+            # Start with None context and chain baggage settings
+            new_baggage = None
+
+            # Add each metadata key-value pair to baggage
+            for key, value in kwargs.items():
+                # Convert value to string and truncate if needed for baggage
+                str_value = str(value)
+                if len(str_value) > 200:
+                    str_value = str_value[:200]
```
Steffen911: We should either apply this centrally or drop it. Just having it on baggage does not make sense. Alternatively, we reject it on runtime if the size exceeds the recommended thresholds.

langfuse/_client/client.py
```diff
@@ -3450,3 +3467,211 @@ def clear_prompt_cache(self) -> None:
...
+
+                baggage_key = f"metadata.{key}"
```
Steffen911: Use `langfuse.metadata.{key}` instead.

langfuse/_client/client.py
```diff
@@ -3450,3 +3467,211 @@ def clear_prompt_cache(self) -> None:
...
+                # Convert value to appropriate type for span attribute
+                if isinstance(value, (str, int, float, bool)):
+                    attr_value = value
+                else:
+                    # For complex types, convert to JSON string
+                    attr_value = json_module.dumps(value)
+                current_span.set_attribute(attr_key, attr_value)
+
+        # Set baggage if requested
+        baggage_tokens = []
```
Steffen911: This should be unnecessary and can be dropped if we attach the baggage only once outside of the kwargs loop.

langfuse/_client/span_processor.py
```diff
@@ -29,6 +32,11 @@
 from langfuse.logger import langfuse_logger
 from langfuse.version import __version__ as langfuse_version
 
+# Context key constants for Langfuse context propagation
+LANGFUSE_CTX_USER_ID = "langfuse.ctx.user.id"
+LANGFUSE_CTX_SESSION_ID = "langfuse.ctx.session.id"
+LANGFUSE_CTX_METADATA = "langfuse.ctx.metadata"
```
Steffen911: Redundant declaration

langfuse/_client/span_processor.py
```diff
@@ -114,6 +122,108 @@ def __init__(
...
+        Args:
+            span: The span that is starting
+            parent_context: The context when the span was created (optional)
+        """
+        if self._is_langfuse_span(span) and not self._is_langfuse_project_span(span):
+            langfuse_logger.debug(
+                f"Security: Span rejected - belongs to project '{span.instrumentation_scope.attributes.get('public_key') if span.instrumentation_scope and span.instrumentation_scope.attributes else None}' but processor is for '{self.public_key}'. "
+                f"This prevents cross-project data leakage in multi-project environments."
+            )
+            return super().on_start(span, parent_context)
```
Steffen911: @hassiebp Do you think it makes sense to have this "Gate" on start as well? Or is it sufficient to use this on end so we drop it here?

langfuse/_client/client.py
```diff
@@ -1667,6 +1672,11 @@ def update_current_trace(
                 span.update(output=response)
             ```
         """
+        warnings.warn(
+            "update_current_trace is deprecated and will be removed in a future version. ",
```
hassiebp: we need to point to docs on what the alternative is now

langfuse/_client/context_propagation.py
```diff
@@ -0,0 +1,244 @@
...
+    for local (same-service) propagation, with optional baggage for cross-service
+    propagation.
+
+    Classes that inherit this mixin gain the ability to create contexts where
+    certain attributes are automatically applied to all spans without manual
+    specification.
+    """
+
+    @_agnosticcontextmanager
+    def session(
```
hassiebp: I think we should simplify this to only one context manager that is simply `trace_attributes` (this could also live directly on the Langfuse class, no need for separate mixin). 

Currently, if the user wants to set session_id, user_id, and metadata they would have to nest three context managers:

```python
with langfuse.session_id('session'):
    with langfuse.user_id('user):
        with langfuse.metadata(*):
            pass
```

We could simplify this to 

```python
with langfuse.trace_attributes(session_id='session', user_id='user', metadata={'key': 1}):
    pass
```

langfuse/_client/context_propagation.py
```diff
@@ -0,0 +1,244 @@
...
+        self, id: str, *, as_baggage: bool = False
```
hassiebp: I think we should rename `as_baggage` to something more immediately descriptive on what the impact is. Something like `propagate_via_http` makes it both clear that this is propagation related and comes with security considerations (users might be unfamiliar with OTEL in general and baggage in particular). We could also consider dropping this entirely and only allow enabling it via env var only as users probably won't dynamically decide in runtime whether to use baggage or not

hassiebp: For propagation, we also need to consider / test the `multiprocessing` case in Python, where users use Celery workers that are separate Python processes to handle background tasks. Trace attributes must be propagated there, too

langfuse/_client/span.py
```diff
@@ -233,6 +238,11 @@ def update_trace(
             tags: List of tags to categorize the trace
             public: Whether the trace should be publicly accessible
         """
+        warnings.warn(
```
hassiebp, line 236: Similar as in client, we need to point out what alternative is

langfuse/_client/span.py
```diff
@@ -52,14 +56,15 @@
...
+from langfuse._client.context_propagation import LangfuseContextPropagationMixin
 
 # Factory mapping for observation classes
 # Note: "event" is handled separately due to special instantiation logic
 # Populated after class definitions
 _OBSERVATION_CLASS_MAP: Dict[str, Type["LangfuseObservationWrapper"]] = {}
 
 
-class LangfuseObservationWrapper:
+class LangfuseObservationWrapper(LangfuseContextPropagationMixin):
```
hassiebp: Mixin not necessary here any longer if we have a single `trace_attributes` context manager on the Langfuse client

langfuse/_client/span_processor.py
hassiebp: The `on_end` is sufficient as that one is handling the actual span export. We can drop this here

langfuse/_client/span_processor.py
```diff
@@ -114,6 +122,108 @@ def __init__(
...
+        propagated_attributes = {}
+
+        # 1. Propagate all baggage keys as span attributes
+        baggage_entries = baggage.get_all(context=current_context)
+        for key, value in baggage_entries.items():
+            # Check if this baggage entry is already present as a span attribute
+            if not hasattr(span.attributes, key) or (
+                span.attributes is not None and span.attributes.get(key) != value
+            ):
+                propagated_attributes[key] = value
```
hassiebp, line 142: We should probably only set the langfuse baggage keys here on all spans to avoid the side effect of propagating all non-langfuse baggage entries to all spans which might be undesired

langfuse/_client/span_processor.py
```diff
@@ -114,6 +122,108 @@ def __init__(
...
+        # Get the current context (use parent_context if available, otherwise current)
+        current_context = parent_context or context_api.get_current()
+
+        # Dictionary to collect span attributes that were propagated
+        propagated_attributes = {}
+
+        # 1. Propagate all baggage keys as span attributes
+        baggage_entries = baggage.get_all(context=current_context)
+        for key, value in baggage_entries.items():
+            # Check if this baggage entry is already present as a span attribute
```
hassiebp: In what cases are the trace attributes already present on_start of the span?

langfuse/_client/context_propagation.py
```diff
@@ -0,0 +1,244 @@
...
+    baggage as otel_baggage_api,
+    context as otel_context_api,
+    trace as otel_trace_api,
+)
+from opentelemetry.util._decorator import _agnosticcontextmanager
+
+# Context key constants for Langfuse context propagation
+LANGFUSE_CTX_USER_ID = "langfuse.ctx.user.id"
+LANGFUSE_CTX_SESSION_ID = "langfuse.ctx.session.id"
+LANGFUSE_CTX_METADATA = "langfuse.ctx.metadata"
```
Steffen911: TODO: We should restrict metadata values to a specific length

langfuse/_client/context_propagation.py
Steffen911: @hassiebp Test langchain

langfuse/_client/client.py
```diff
@@ -350,6 +360,108 @@ def start_span(
...
+        if metadata is not None:
+            # Truncate values with size > 200 to 200 characters and emit warning including the ky
+            for k, v in metadata.items():
+                if not isinstance(v, str):
+                    # Ignore unreachable mypy warning as this runtime guard should make sense either way
+                    warnings.warn(  # type: ignore[unreachable]
+                        f"Metadata values must be strings, got {type(v)} for key '{k}'"
+                    )
+                    del metadata[k]
+                if len(v) > 200:
```
philipbjorge: We ran into an error here in production where `v` is `None` and this length check failed.

```
  File "/app/documents_parser/telemetry/middleware.py", line 41, in dispatch
    with client.with_attributes(
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/opentelemetry/util/_decorator.py", line 61, in __enter__
    return next(self.gen)  # type: ignore
           ^^^^^^^^^^^^^^
  File "/app/.venv/lib/python3.12/site-packages/langfuse/_client/client.py", line 440, in with_attributes
    if len(v) > 200:
       ^^^^^^
TypeError: object of type 'NoneType' has no len()
```

philipbjorge: It's a bug on our side, but I think the package should probably warn and fail gracefully.

langfuse/_client/span.py
maxdeichmann: Can we also deprecate this in the SDK reference?

langfuse/_client/client.py
```diff
@@ -211,8 +219,10 @@ def __init__(
         additional_headers: Optional[Dict[str, str]] = None,
         tracer_provider: Optional[TracerProvider] = None,
     ):
-        self._host = host or cast(
-            str, os.environ.get(LANGFUSE_HOST, "https://cloud.langfuse.com")
+        self._host = (
+            host
+            if host is not None
+            else os.environ.get(LANGFUSE_HOST, "https://cloud.langfuse.com")
         )
```
maxdeichmann, line 229: The original version is more readableand to my understanding they do the same. If this is true, should we revert this?

langfuse/_client/client.py
```diff
@@ -350,6 +360,108 @@ def start_span(
...
+    ) -> Generator[None, None, None]:
+        """Creates a context manager that propagates the given attributes to all spans created within the context.
+
+        Args:
+            session_id (str): Session identifier.
+            user_id (str): User identifier.
+            metadata (dict): Additional metadata to associate with all spans in the context. Values must be strings and are truncated to 200 characters.
+            as_baggage (bool, optional): If True, stores the values in OpenTelemetry baggage
+                for cross-service propagation. If False, stores only in local context for
+                current-service propagation. Defaults to False.
```
maxdeichmann, line 381: We should add a sentence here that this will be added to the headers of all outgoing HTTP requests. I would add the warning from below up here as not everyone is reading everything.

langfuse/_client/client.py
```diff
@@ -350,6 +360,108 @@ def start_span(
...
+                current_span.set_attribute("user.id", user_id)
+            if as_baggage:
+                current_context = otel_baggage_api.set_baggage(
+                    "user.id", user_id, current_context
+                )
+
+        # Process metadata
+        if metadata is not None:
+            # Truncate values with size > 200 to 200 characters and emit warning including the ky
+            for k, v in metadata.items():
```
maxdeichmann: Should we also have an overall limit fo e.g. 4k characters?

Steffen911: @maxdeichmann IMO that would be complicated to enforce and eventually, this is an issue that will surface quickly on the client side. I would highlight it in the documentation, but don't complicate the happy path further.

langfuse/_client/client.py
```diff
@@ -350,6 +360,108 @@ def start_span(
             status_message=status_message,
         )
 
+    @_agnosticcontextmanager
+    def with_attributes(
+        self,
+        session_id: Optional[str] = None,
+        user_id: Optional[str] = None,
+        metadata: Optional[dict[str, str]] = None,
```
Steffen911: Should we highlight somehow that this metadata is only to be used if it should be set on _all_ children? I'm wondering whether people understand and make use of the difference for setting metadata on _current_ span vs _all_ spans?

maxdeichmann: Yes, I agree with the sentiment! Options:
- with_trace_attributes
- with_context_attributes
- with_propagated_attributes

...
Lets quickly talk sync about this. Nailing this is key.

maxdeichmann:
@Steffen911 lets talk about function naming of the new functions and then we are good to go here.

langfuse/_client/client.py
maxdeichmann: Well the risk is here that our users run into failing APIs as the header is too large to be processed. I thought we could add a simple logic such as: keeping a count of added characters to the header. If threshhold exceeded, stop adding key/values to the metadata.

langfuse/_client/client.py
hassiebp: yes, they are both equivalent

langfuse/_client/client.py
hassiebp: We should think about this from the perspective on how this is consumed, namely via context manager. so the 'with' is already there.

```python
from langfuse import get_client

langfuse = get_client()

...

with langfuse.correlation_context(session_id='my-session'):
    # do stuff
    pass
```

I prefer `correlation_context` or `set_correlation_context` as it makes sure that 
- this is not necessarily related to a 'trace' 
- this has some propagation included ('context')
- is used for aggregating / correlating multiple spans ('correlation')

maxdeichmann: * propagation_context
* propagated_context
* distributed_context
* trace_scope
* trace_context
* shared_attributes

PR #1566, opened 2026-03-18
fix(observe): Handle asyncio.CancelledError in exception blocks

jinohkang-theori:
Fixes langfuse/langfuse#12650

This catches and records the async task being cancelled—either directly or because it timed out—which will raise an asyncio.CancelledError to stop execution.

Also fix sync wrappers as well. Note that asyncio.CancelledError can cross async-sync boundaries, such as `run_coroutine_threadsafe().result()` or a sync helper function that re-raises cancellation that was previously suppressed/deferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 ## Installation
 
 > [!IMPORTANT]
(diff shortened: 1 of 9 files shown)


hassiebp:
Thanks a lot for your contribution, @jinohkang-theori !

PR #1572, opened 2026-03-20
fix: coerce non-string metadata values instead of dropping them

### LINKED ISSUE #1571: bug: propagate_attributes drops non-string metadata instead of coercing (breaks LangGraph integration)
## Description

The v4 migration guide states that propagated metadata non-string values are "automatically coerced to strings." However, the actual implementation in `propagation.py` drops non-string values with a warning instead of coercing them.

This causes noisy warnings when using LangGraph with the Langfuse CallbackHandler, since LangGraph injects internal metadata with non-string values into the `RunnableConfig` at runtime:

- `langgraph_step` → `int`
- `langgraph_triggers` → `list`
- `langgraph_path` → `tuple`
- Custom configurable values like `max_search_results` → `int`

## Expected behavior

Per the [v3 → v4 migration guide](https://langfuse.com/docs/observability/sdk/upgrade-path/python-v3-to-v4):

> propagated metadata: now `dict[str, str]` with values limited to 200 characters (was `Any`). **Non-string values are automatically coerced to strings.**

Non-string metadata values should be coerced to strings (e.g., via `str(value)` or `json.dumps(value)`) rather than dropped.

## Actual behavior

`_validate_string_value` in `propagation.py:419-424` drops non-string values and logs a warning:

```
WARNING:langfuse:Propagated attribute 'metadata.langgraph_step' value is not a string. Dropping value.
WARNING:langfuse:Propagated attribute 'metadata.langgraph_triggers' value is not a string. Dropping value.
WARNING:langfuse:Propagated attribute 'metadata.langgraph_path' value is not a string. Dropping value.
```

These warnings appear on every LangGraph agent invocation since LangGraph adds these metadata keys automatically.

## Suggested fix

In `propagation.py`, coerce non-string values instead of dropping them:

```python
def _validate_string_value(*, value: str, key: str) -> bool:
    if not isinstance(value, str):
        # Coerce to string instead of dropping
        return True  # after converting: value = str(value)
```

Or handle this at the metadata processing level in `_propagate_attributes` by converting non-string values before validation:

```python
if metadata is not None:
    validated_metadata: Dict[str, str] = {}
    for key, value in metadata.items():
        str_value = str(value) if not isinstance(value, str) else value
        if _validate_string_value(value=str_value, key=f"metadata.{key}"):
            validated_metadata[key] = str_value
```

## Environment

- `langfuse` Python SDK version: 4.0.0
- `langgraph` version: >= 1.0.0
- Python version: 3.12

## Reproduction

```python
from langfuse.langchain import CallbackHandler
from langgraph.graph import StateGraph, MessagesState, START

handler = CallbackHandler()

# Any LangGraph agent will trigger the warnings because LangGraph
# injects langgraph_step (int), langgraph_triggers (list), langgraph_path (tuple)
# into the RunnableConfig metadata during execution.
builder = StateGraph(MessagesState)
# ... build graph ...
graph = builder.compile()

for chunk in graph.stream(
    {"messages": [{"role": "user", "content": "hello"}]},
    config={"callbacks": [handler]},
    stream_mode="values",
):
    pass
# Warnings appear in logs for every graph step
```

nicoyoung:
## Summary

- Coerces non-string metadata values to strings via `str()` before validation in `propagate_attributes()`, matching the [v4 migration guide](https://langfuse.com/docs/observability/sdk/upgrade-path/python-v3-to-v4) which states non-string values are "automatically coerced to strings"
- Fixes noisy warnings when using LangGraph with the Langfuse CallbackHandler, since LangGraph injects non-string metadata (`langgraph_step=int`, `langgraph_triggers=list`, `langgraph_path=tuple`) into the RunnableConfig at runtime

## Change

In `langfuse/_client/propagation.py`, added `str()` coercion before `_validate_string_value` for metadata values:

```python
for key, value in metadata.items():
    if not isinstance(value, str):
        value = str(value)
    if _validate_string_value(value=value, key=f"metadata.{key}"):
        validated_metadata[key] = value
```

Fixes #1571

---

<h3>Greptile Summary</h3>

This PR fixes noisy warnings when using the Langfuse `CallbackHandler` with LangGraph by coercing non-string metadata values to strings (via `str()`) in `_propagate_attributes()` instead of dropping them, matching the documented v4 migration guide behaviour.

- **Coercion logic** (`propagation.py` lines 258–259): The change is minimal and correct — non-string values are coerced with `str()` before the existing `_validate_string_value` call, so the 200-character length guard still applies after coercion.
- **Type hint inconsistency**: Both `propagate_attributes()` (line 79) and `_propagate_attributes()` (line 213) still declare `metadata: Optional[Dict[str, str]]`. Since the runtime now accepts and coerces non-string values, these should be updated to `Optional[Dict[str, Any]]` to avoid spurious type errors for callers passing LangGraph-style mixed dicts.
- **Stale docstring**: The `propagate_attributes` docstring (line 112) still warns that non-string values *"will be dropped with a warning"*, and the `Note:` section (lines 188–190) still says values *"must be strings"*. Both statements are now incorrect and should be updated.

<h3>Confidence Score: 4/5</h3>

- Safe to merge — the coercion logic is correct and the existing length guard still applies, but type hints and docstrings need a follow-up update.
- The one-line fix is well-targeted and aligns with the migration guide. The only concerns are a stale type annotation and outdated docstring, neither of which affect runtime correctness.
- langfuse/_client/propagation.py — type hint on `metadata` parameter and the docstring both still describe the old drop-on-non-string behaviour.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| langfuse/_client/propagation.py | Adds `str()` coercion for non-string metadata values before validation in `_propagate_attributes()`; the fix itself is correct and safe, but the `metadata` type hint (`Dict[str, str]`) and the docstring still reflect the old drop-on-non-string behaviour and need updating. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["propagate_attributes() called\nwith metadata dict"] --> B{metadata is not None?}
    B -- No --> G[Skip metadata processing]
    B -- Yes --> C[Iterate over metadata key-value pairs]
    C --> D{value is a string?}
    D -- Yes --> E["_validate_string_value(value, key)"]
    D -- No --> F["value = str(value)  ← NEW\n(coerce non-string to string)"]
    F --> E
    E --> H{Valid?\nlength ≤ 200 chars}
    H -- Yes --> I[Add to validated_metadata]
    H -- No --> J[Log warning, drop value]
    I --> K[_set_propagated_attribute\nwith validated_metadata]
    J --> C
    K --> L[Set in OTel context & current span]
```

<sub>Last reviewed commit: ["fix: coerce non-stri..."](https://github.com/langfuse/langfuse-python/commit/3ea00c2895d403fb1514af0046b8a66d4d7ab0e0)</sub>

<sub>(2/5) Greptile learns from your feedback when you react with thumbs up/down!</sub>


nicoyoung:
Fixed in https://github.com/langfuse/langfuse-python/pull/1709

PR #1577, opened 2026-03-26
fix: add depth limit to EventSerializer to prevent hangs on complex objects

BKDDFS:
## Summary

`EventSerializer.default()` recursively traverses `__dict__` and `__slots__` of arbitrary objects without a depth limit. When `@observe()` captures function arguments containing objects like `google.genai.Client` (which hold aiohttp sessions, connection pools, and threading locks), `json.dumps` can block indefinitely.

### Root cause

`_serialize()` in `create_span_attributes` calls `json.dumps(obj, cls=EventSerializer)`. The serializer's `__dict__` fallback recurses into every attribute of every object. For objects like `aiohttp.ClientSession` → `TCPConnector` → internal locks, this either hangs (lock contention / blocking attribute access) or enters extremely deep recursion.

### Fix

Add a `_MAX_DEPTH = 20` depth counter to `EventSerializer`:
- `default()` increments/decrements `_depth` on each call
- When `_depth >= _MAX_DEPTH`, returns `<TypeName>` placeholder instead of recursing
- Serialization logic moved to `_default_inner()` — no behavior changes
- `encode()` resets `_depth` at the start of each serialization

### Workaround (current)

`@observe(capture_input=False)` — disables automatic input capture entirely.

### Tests

3 new tests:
- `test_deeply_nested_object_does_not_hang` — objects with connection pools/locks
- `test_max_depth_returns_type_name` — `__dict__` deep nesting truncation
- `test_deeply_nested_slots_object_is_truncated` — `__slots__` deep nesting truncation

All 21 tests pass (18 existing + 3 new).

---

<h3>Greptile Summary</h3>

This PR fixes a real and reproducible hang in `EventSerializer` by introducing a `_MAX_DEPTH = 20` counter that short-circuits recursion when serializing deeply nested objects (e.g. `google.genai.Client` holding aiohttp session internals). The core mechanism — incrementing `_depth` in `default()` and decrementing it in a `finally` block — is correct and well-structured.\n\n**Key changes:**\n- `default()` is now a thin depth-guard that delegates to `_default_inner()`, preserving all existing serialization behavior unchanged.\n- `encode()` resets `_depth = 0` before each top-level call, which is a good defensive measure.\n- Three new tests cover the main scenarios: hang prevention, `__dict__` depth truncation, and `__slots__` depth truncation.\n\n**Issues found:**\n- The `__slots__` branch at line 153–156 calls `self.default({...})` for the intermediate dict it constructs. Because `default()` also increments `_depth`, each `__slots__` object consumes **2 depth units** per level instead of 1, cutting the effective limit from 20 to ~10 for `__slots__`-based objects. Changing that single call to `_default_inner({...})` would make behavior consistent.\n- The `test_deeply_nested_slots_object_is_truncated` assertion bound (`_MAX_DEPTH + 5 = 25`) is too loose to catch regressions in truncation depth for the slots case.

<h3>Confidence Score: 4/5</h3>

Safe to merge — the hang is fixed and no existing behavior is regressed; the remaining issues are a minor depth inconsistency for `__slots__` objects and a loose test assertion.

The fix correctly solves the stated problem (infinite recursion / hang on complex objects) and all 21 tests pass. The only issues are a double-depth-counting side-effect in the `__slots__` path (effective limit ~10 instead of 20) and an overly permissive test assertion — neither affects the primary fix or causes correctness failures in normal usage.

`langfuse/_utils/serializer.py` lines 153–156 (`__slots__` handling) and the corresponding test assertion at `tests/test_serializer.py:267`.

<h3>Important Files Changed</h3>

| Filename | Overview |
|----------|----------|
| langfuse/_utils/serializer.py | Adds `_MAX_DEPTH = 20` depth limiter to prevent infinite recursion/hangs; logic is correct for `__dict__`-based objects but `__slots__` objects incur double depth consumption (consuming 2 units per level instead of 1), halving their effective limit to ~10. |
| tests/test_serializer.py | Adds 3 new tests covering hang prevention, `__dict__` truncation, and `__slots__` truncation; the slots test has an overly loose assertion (`_MAX_DEPTH + 5`) that doesn't tightly verify the truncation depth. |

</details>

<h3>Flowchart</h3>

```mermaid
flowchart TD
    A["encode(obj)"] --> B["_depth = 0\nseen.clear()"]
    B --> C["default(obj)"]
    C --> D{{"_depth >= _MAX_DEPTH?"}}
    D -- Yes --> E["return '<TypeName>'"]
    D -- No --> F["_depth += 1\n_default_inner(obj)"]
    F --> G{{"Type of obj?"}}
    G -- "datetime / UUID / bytes / etc." --> H["Return primitive"]
    G -- "dict" --> I["self.default(k)\nself.default(v)\nfor each item"]
    G -- "list / Sequence" --> J["self.default(item)\nfor each item"]
    G -- "__slots__" --> K["self.default(dict of slots)\n⚠️ extra depth unit"]
    G -- "__dict__" --> L{{"id(obj) in seen?"}}
    L -- Yes --> M["return TypeName (circular ref)"]
    L -- No --> N["seen.add(id)\nself.default(v)\nfor each attr\nseen.remove(id)"]
    G -- "other" --> O["return '<TypeName>'"]
    I --> P["_depth -= 1\n(finally)"]
    J --> P
    K --> P
    N --> P
    H --> P
    M --> P
    O --> P
    P --> Q["return result to encode()"]
    Q --> R["super().encode(result)"]
```

<details><summary><h3>Comments Outside Diff (1)</h3></summary>

1. `langfuse/_utils/serializer.py`, line 153-156 ([link](https://github.com/langfuse/langfuse-python/blob/a895aed67a125d51112060701d3e467d6807dacf/langfuse/_utils/serializer.py#L153-L156)) 

   <a href="#">P2</a> **`__slots__` path consumes double the depth budget per object level**

   Because the `__slots__` branch calls `self.default({...})` for the intermediate dict, each `__slots__` object consumes **2** depth units (one for the object itself, one for the constructed dict), whereas a `__dict__`-based object consumes only **1**. In practice this halves the effective depth limit for `__slots__` objects: they truncate at approximately 10 levels of nesting rather than the intended 20.

   Tracing for a chain of `SlotLevel` objects:
   - `default(SlotLevel_N)` → `_depth` = 2*(N-1)
   - `default({"child": SlotLevel_{N+1}})` → `_depth` = 2*(N-1)+1
   - Cutoff: 2*(N-1) ≥ 20 → N ≥ 11 (truncates at level 11, not 21)

   A lightweight fix is to call `_default_inner` directly for the constructed dict rather than routing it back through `default()`:

   ```python
   if hasattr(obj, "__slots__"):
       return self._default_inner(
           {slot: getattr(obj, slot, None) for slot in obj.__slots__}
       )
   ```

   This avoids the redundant depth check and gives `__slots__` objects the same effective depth limit as `__dict__` objects.

</details>

<sub>Reviews (1): Last reviewed commit: ["fix: add depth limit to EventSerializer ..."](https://github.com/langfuse/langfuse-python/commit/a895aed67a125d51112060701d3e467d6807dacf) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=26447345)</sub>

<sub>(2/5) Greptile learns from your feedback when you react with thumbs up/down!</sub>


BKDDFS:
friendly bump @hassiebp

BKDDFS:
Another friendly bump @hassiebp / @marcklingen. This is open for ~50 days now with no review.

The hang causes real impact in our prod: we ship `@observe(capture_input=False)` as a workaround on every endpoint that uses clients like `google.genai.Client` / aiohttp-backed clients, which means traces lose input data entirely on those flows. Switching to a depth-limited serializer would let us re-enable input capture.

Happy to:
- rebase to resolve the conflict
- address Greptile's `__slots__` double-depth nit (one-line fix: route the slots dict through `_default_inner` instead of `default`)

if either is blocking the review.

hassiebp:
@claude review

hassiebp:
@BKDDFS Thanks a lot for your contribution! This looks good to me - could you kindly resolve the merge conflicts for this to get merged?

BKDDFS:
Addressed claude comments with regression tests and resolved conflicts @hassiebp

hassiebp:
Thanks for your contribution @BKDDFS !

PR #1646, opened 2026-05-07
feat(tracing): add export-stage OTEL span masking and media detection

### LINKED ISSUE #11931: Python SDK masking should allow batch processing of spans prior to export
Currently users can register a mask function that is invoked on span creations to redact PII from the created OTEL spans.

This redaction happens (a) sync at span creation time and (b) on a span by span basis.

For users that rely on IO bound span masking, it would be great to move masking 

(a) to a background thread to not block the main thread
(b) allow batch processing in masking to limit network calls made to a masking backend

hassiebp:
## Summary

Adds an export-stage transformation layer for spans before they are sent to the downstream OTLP exporter.

- adds the public `mask_otel_spans` callback on `Langfuse(...)`
- adds the batch-shaped public contract: `MaskOtelSpansParams`, `MaskOtelSpansResult`, `OtelSpanIdentifier`, `OtelSpanData`, and `OtelSpanPatch`
- wires `mask_otel_spans` through the resource manager into the Langfuse span processor
- runs export-stage media detection over exported span attributes before masking
- prefilters serialized string attributes before JSON parsing so JSON-looking strings without supported media hints are left unchanged cheaply
- applies sparse keyed span patches to post-media OTEL span attribute snapshots before export
- keeps existing synchronous `mask` behavior unchanged
- hardens media detection for Gemini `inline_data` / `inlineData` payloads

## Why

Attribute masking and media detection currently happen when Langfuse SDK spans set attributes. That leaves third-party OTEL spans flowing through the Langfuse span processor without equivalent export-stage handling. This PR adds a batch-native export-path hook while preserving existing Langfuse SDK masking behavior.

## Behavior

- `mask_otel_spans` runs only after existing export filters accept a span.
- Media detection runs before `mask_otel_spans`.
- Direct base64 data URI string attributes are processed without JSON parsing.
- JSON-looking string attributes are parsed only if they contain supported media-shape hints such as `data:`, `inline_data`, `inlineData`, `media_type`, `mime_type`, or `mimeType` plus `data`.
- `mask_otel_spans` receives the whole OTEL export batch as `params.spans`, keyed by `OtelSpanIdentifier(trace_id, span_id)`.
- `MaskOtelSpansResult.span_patches` is sparse: omitted spans are exported unchanged.
- `OtelSpanPatch` can delete exact attribute keys and set attribute values; delete runs before set, so set wins.
- Callback exceptions or invalid outer batch results drop the whole export batch.
- Patches for unknown span identifiers drop the whole export batch.
- Invalid patch objects for known spans drop only those spans.
- Invalid returned set values delete only that affected attribute and log the value type, not the value.
- Span events and links are not transformed.

## Validation

- `uv run --frozen ruff check .`
- `uv run --frozen pytest tests/unit/test_mask_otel_spans.py tests/unit/test_media_manager.py -q`
- `uv run --frozen pytest tests/unit/test_otel.py tests/unit/test_additional_headers_simple.py -q`
- `uv run --frozen pytest -n auto --dist worksteal tests/unit`

---

This PR adds an export-stage transformation layer (`LangfuseTransformingSpanExporter`) that runs between the OTel batch span processor and the downstream OTLP exporter, enabling media detection and attribute masking on third-party spans that bypass the existing SDK-level `mask` hook.

- Introduces `mask_otel_spans` as a new public `Langfuse(...)` parameter, wired through the resource manager into `LangfuseSpanProcessor`, and exposes a batch-native public contract (`MaskOtelSpansParams`, `MaskOtelSpansResult`, `OtelSpanPatch`, etc.) in `langfuse.types`.
- Extends `MediaManager._find_and_process_media` with a `fail_open` flag and adds Gemini `inline_data`/`inlineData` detection; applies media replacement at export time with a substring prefilter to avoid JSON parsing on non-media strings.
- The OTEL tracer initialization in `resource_manager.py` was moved after the media manager setup so `media_manager` is available when the span processor is constructed.
</details>

Safe to merge; all changes are additive and opt-in, existing mask and media paths are untouched.

The change is well-structured and defensively written, but cloned spans lose their `dropped_attributes_count` metadata — downstream OTLP consumers would see 0 even for spans that originally exceeded the attribute limit.

langfuse/_client/span_exporter.py — specifically the _clone_span method and attribute wrapping strategy.
</details>

```mermaid
sequenceDiagram
    participant App as Application
    participant BSP as BatchSpanProcessor
    participant LTS as LangfuseTransformingSpanExporter
    participant MM as MediaManager
    participant MF as mask_otel_spans callback
    participant OE as OTLPSpanExporter

    App->>BSP: span ends (on_end)
    BSP->>LTS: export(batch of ReadableSpans)

    loop for each span
        LTS->>MM: "_find_and_process_media(attributes, fail_open=True)"
        MM-->>LTS: post-media attribute dict
    end

    alt mask_otel_spans configured
        LTS->>MF: MaskOtelSpansParams(spans)
        MF-->>LTS: MaskOtelSpansResult or None

        alt exception or invalid result
            LTS-->>BSP: SUCCESS (batch dropped)
        else patch for unknown identifier
            LTS-->>BSP: SUCCESS (batch dropped)
        else valid patches
            loop for each span
                LTS->>LTS: apply OtelSpanPatch (delete then set)
            end
        end
    end

    LTS->>LTS: _clone_span(span, patched_attributes)
    LTS->>OE: export(cloned ReadableSpans)
    OE-->>LTS: SpanExportResult
    LTS-->>BSP: SpanExportResult
```
</details>

`````markdown
Fix the following 2 code review issues. Work through them one at a time, proposing concise fixes.

---

### Issue 1 of 2
langfuse/_client/span_exporter.py:579-596
`_clone_span` converts `span.attributes` (a `BoundedAttributes`) to a plain `dict`. The OTel SDK's `ReadableSpan.dropped_attributes` property guards with `isinstance(self._attributes, BoundedAttributes)` and returns `0` for any other type, so the cloned span will always export `dropped_attributes_count=0` even when the original span had attributes silently dropped by the bounds limit. This loses span fidelity in the OTLP payload for any span that hit the attribute limit.

```suggestion
    @staticmethod
    def _clone_span(
        *, span: ReadableSpan, attributes: Dict[str, AttributeValue]
    ) -> ReadableSpan:
        return ReadableSpan(
            name=span.name,
            context=span.context,
            parent=span.parent,
            resource=span.resource,
            attributes=BoundedAttributes(
                maxlen=None,
                attributes=attributes,
                immutable=True,
                max_value_len=None,
            ),
            events=span.events,
            links=span.links,
            kind=span.kind,
            status=span.status,
            start_time=span.start_time,
            end_time=span.end_time,
            instrumentation_scope=span.instrumentation_scope,
        )
```

### Issue 2 of 2
langfuse/_client/span_exporter.py:661-669
**Prefilter false-positive on `data:` substrings in URLs** — `_may_contain_serialized_media` triggers on any string containing the literal `data:` (e.g., `"https://api.example.com/data:types"`), causing unnecessary `json.loads` parsing before the string is returned unchanged. No data is lost, but large attribute strings with `data:` in a URL incur a redundant parse on every export.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["test(tracing): cover export-stage span m..."](https://github.com/langfuse/langfuse-python/commit/82a8eb93efdbeb812c57c6d4a6588a3d3f9fefe4) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31185632)</sub>

### CODE REVIEW

File: langfuse/_client/client.py [hunk 1]
```diff
@@ -170,6 +176,7 @@ class Langfuse:
         media_upload_thread_count (Optional[int]): Number of background threads for handling media uploads. Defaults to 1. Can also be set via LANGFUSE_MEDIA_UPLOAD_THREAD_COUNT environment variable.
         sample_rate (Optional[float]): Sampling rate for traces (0.0 to 1.0). Defaults to 1.0 (100% of traces are sampled). Can also be set via LANGFUSE_SAMPLE_RATE environment variable.
         mask (Optional[MaskFunction]): Function to mask sensitive data in traces before sending to the API.
+        span_mask (Optional[SpanMaskFunction]): Synchronous export-stage hook that can set or delete attributes on every exported span before it is sent to Langfuse. If the hook raises or returns an invalid result, the span is dropped. During flush and shutdown, this hook may run on the caller thread.
```
hassiebp: make comment clear that this runs on background thread. make it also clear what difference is to `mask`, esp. that this is only applied to spans sent to Langfuse, not to third-party observability backends.

langfuse/_client/client.py
```diff
@@ -350,6 +360,7 @@ def __init__(
             span_exporter=span_exporter,
         )
         self._mask = self._resources.mask
+        self._span_mask = self._resources.span_mask
```
hassiebp: is this used anywhere? if not, drop it

langfuse/_client/client.py
```diff
@@ -170,6 +176,7 @@ class Langfuse:
         media_upload_thread_count (Optional[int]): Number of background threads for handling media uploads. Defaults to 1. Can also be set via LANGFUSE_MEDIA_UPLOAD_THREAD_COUNT environment variable.
         sample_rate (Optional[float]): Sampling rate for traces (0.0 to 1.0). Defaults to 1.0 (100% of traces are sampled). Can also be set via LANGFUSE_SAMPLE_RATE environment variable.
         mask (Optional[MaskFunction]): Function to mask sensitive data in traces before sending to the API.
+        mask_otel_spans (Optional[MaskOtelSpansFunction]): Synchronous export-stage hook that receives a batch of OpenTelemetry span snapshots and can return sparse attribute patches before spans are sent to Langfuse. If the hook raises or returns an invalid batch result, the whole export batch is dropped. During flush and shutdown, this hook may run on the caller thread.
```
hassiebp: make clearer:
- difference to `mask` (running in background, only applying to spans sent to Langfuse...)
- batching is dependent on export batch settings
...

langfuse/_client/client.py
```diff
@@ -226,6 +233,7 @@ class Langfuse:
 
     _resources: Optional[LangfuseResourceManager] = None
     _mask: Optional[MaskFunction] = None
+    _mask_otel_spans: Optional[MaskOtelSpansFunction] = None
```
hassiebp: is this used anywhere? if not, remove

langfuse/_client/span_exporter.py
```diff
@@ -0,0 +1,501 @@
...
+
+        if not transformed_spans:
+            return SpanExportResult.SUCCESS
+
+        return self._exporter.export(transformed_spans)
+
+    def shutdown(self) -> None:
+        self._exporter.shutdown()
+
+    def force_flush(self, timeout_millis: int = 30000) -> bool:
```
hassiebp, line 143: why are we setting a separate default here? can we not rely on the default provided by self._exporter?

langfuse/_client/span_exporter.py
```diff
@@ -0,0 +1,501 @@
...
+
+        for key, value in attributes.items():
+            try:
+                processed_attributes[key] = self._process_media_attribute_value(
+                    span=span,
+                    attribute_key=key,
+                    value=value,
+                )
+            except Exception as error:
+                langfuse_logger.warning(
```
hassiebp, line 162: will logging work since this is running in background thread?

langfuse/_client/span_exporter.py
```diff
@@ -0,0 +1,582 @@
...
+
+    def _process_media_string(
+        self,
+        *,
+        span: ReadableSpan,
+        attribute_key: str,
+        value: str,
+    ) -> str:
+        media_manager = cast(MediaManager, self._media_manager)
+        trace_id = _get_trace_id(span)
```
hassiebp: trace_id or observation_id should never be empty string if span.context is None. if any of those are unavailable, fail the media processing

langfuse/_client/span_exporter.py
```diff
@@ -0,0 +1,582 @@
...
+        *,
+        exporter: SpanExporter,
+        media_manager: Optional[MediaManager],
+        mask_otel_spans: Optional[MaskOtelSpansFunction],
+    ) -> None:
+        self._exporter = exporter
+        self._media_manager = media_manager
+        self._mask_otel_spans = mask_otel_spans
+
+    def export(self, spans: Sequence[ReadableSpan]) -> SpanExportResult:
```
hassiebp, line 108: verify that failing `export` method does not kill the background export thread but exceptions are gracefully handled

PR #1657, opened 2026-05-18
fix(serializer): detect cycles in dict/list/Sequence/__slots__ branches

### LINKED ISSUE #1655: EventSerializer.default recurses forever on cyclic dict/list/__slots__ graphs (blocks asyncio event loop)
# langfuse-python — `EventSerializer.default` recurses forever on cyclic graphs (dict/list/`__slots__` branches missing cycle detection)

**Target repo:** https://github.com/langfuse/langfuse-python
**Affects:** Every published version including the latest 4.6.1 (and all of 3.x)
**Severity:** High — silently blocks the asyncio event loop in any host process

---

## Summary

`langfuse._utils.serializer.EventSerializer.default` has cycle detection on
the `__dict__` branch only. The `dict`, `list`, and `__slots__` branches
recurse without checking `self.seen`, so any object graph that contains a
cycle through one of those types makes the serializer recurse forever.

The `RecursionError` is eventually raised inside Python's GC machinery
("Exception ignored in: `<function WeakValueDictionary.__init__.<locals>.remove>`"),
which Python swallows. By that point the asyncio event loop has been
GIL-starved for minutes.

## Affected source

`langfuse/_utils/serializer.py` in 4.6.1 (line numbers approximate, same
structure in every published version):

```python
# Lines 129-132 — dict branch: NO cycle check
if isinstance(obj, dict):
    return {self.default(k): self.default(v) for k, v in obj.items()}

# Lines 134-136 — list branch: NO cycle check
if isinstance(obj, list):
    return [self.default(item) for item in obj]

# Lines 141-144 — __slots__ branch: NO cycle check
if hasattr(obj, "__slots__"):
    return self.default(
        {slot: getattr(obj, slot, None) for slot in obj.__slots__}
    )

# Lines 145-156 — __dict__ branch: HAS cycle check (the only one)
elif hasattr(obj, "__dict__"):
    obj_id = id(obj)
    if obj_id in self.seen:
        return type(obj).__name__   # <-- correctly breaks the cycle
    else:
        self.seen.add(obj_id)
        result = {k: self.default(v) for k, v in vars(obj).items()}
        self.seen.remove(obj_id)
        return result
```

## Minimal repro (10 lines)

```python
from langfuse._utils.serializer import EventSerializer

d = {"a": 1}
d["self"] = d  # cycle through the dict branch

# This hangs the interpreter at the recursion limit
# (RecursionError eventually raised inside a GC finalizer
# and SILENTLY SWALLOWED). With cycle detection, this
# would return something like '{"a": 1, "self": "<cycle:dict>"}'
EventSerializer().encode(d)
```

Same bug reproduces via `list`:

```python
lst = [1, 2, 3]
lst.append(lst)
EventSerializer().encode(lst)  # also hangs
```

And via `__slots__`-only objects with a self-referential slot value.

## Production impact

Captured in production 2026-05-15. A worker process using the Langfuse
`@observe` decorator over an asyncio FastAPI app experienced seven
consecutive event-loop hangs over six hours — every ~60 minutes,
coinciding with an OAuth client_credentials token refresh that emitted
a log record with a cyclic payload through Langfuse's logging
integration. Each hang locked the asyncio loop until an external
watchdog killed the container (~5 minutes of dead time per hang).

`faulthandler.dump_traceback_later(90)` captured the MainThread cycling
between the dict, list, and `__slots__` branches of `serializer.py` for
the full ~150 frames the dump cap allows. Stacks available on request.

## Suggested fix

Apply the same `seen`-id pattern that already exists on the `__dict__`
branch to the other three branches. Sketch:

```python
def default(self, obj):
    # ... unchanged for BaseModel, Path, Serializable, int, str, etc ...

    if isinstance(obj, (tuple, set, frozenset)):
        return list(obj)

    if isinstance(obj, dict):
        obj_id = id(obj)
        if obj_id in self.seen:
            return f"<cycle:{type(obj).__name__}>"
        self.seen.add(obj_id)
        try:
            return {self.default(k): self.default(v) for k, v in obj.items()}
        finally:
            self.seen.discard(obj_id)

    if isinstance(obj, list):
        obj_id = id(obj)
        if obj_id in self.seen:
            return f"<cycle:{type(obj).__name__}>"
        self.seen.add(obj_id)
        try:
            return [self.default(item) for item in obj]
        finally:
            self.seen.discard(obj_id)

    # Same pattern for the __slots__ branch.
```

`try/finally` ensures the id is removed on exit so the same object
appearing in a sibling subtree (not a cycle) doesn't get spuriously
marked as a cycle.

## Workaround (downstream)

Until a fix lands, a downstream consumer can monkey-patch
`EventSerializer.default` at import time:

```python
from langfuse._utils.serializer import EventSerializer

_original_default = EventSerializer.default

def _cycle_safe_default(self, obj):
    is_dict_or_list = isinstance(obj, (dict, list))
    is_slots_only = hasattr(obj, "__slots__") and not hasattr(obj, "__dict__")
    if is_dict_or_list or is_slots_only:
        obj_id = id(obj)
        if obj_id in self.seen:
            return f"<cycle:{type(obj).__name__}>"
        self.seen.add(obj_id)
        try:
            return _original_default(self, obj)
        finally:
            self.seen.discard(obj_id)
    return _original_default(self, obj)

EventSerializer.default = _cycle_safe_default
```

(Idempotent if guarded by an attribute flag.)

## Environment

- langfuse 4.6.1 (Python wheel from PyPI)
- Python 3.13.13
- Linux/glibc (Docker, `python:3.13-slim`)
- httpx 0.x via langfuse's own deps

Happy to provide additional context / stacks / a draft PR if useful.

wtfashwin:
## Summary

Fixes #1655.

`EventSerializer.default` only guarded the `__dict__` branch against cyclic object graphs. Cycles through the `dict`, `list`, `Sequence`, or `__slots__` branches recursed until the resulting `RecursionError` was swallowed inside Python's GC machinery, leaving the asyncio event loop GIL-starved for minutes (a real production symptom — repro and stack details on the issue).

The same `id()`-keyed `seen` pattern is now applied to the four uncovered branches. Each guard is wrapped in `try` / `finally` with `set.discard`, so a shared object reachable from two sibling subtrees (a DAG, not a cycle) is **not** mis-marked. The pre-existing `__dict__` branch behaviour and cycle-marker format are unchanged — the change is purely additive.

Marker format on the new branches is `"<cycle:<Type>>"` (as suggested in the issue body); the existing `__dict__` marker stays as bare `type(obj).__name__` to preserve back-compat.

### Why this matters
- Silent: the `RecursionError` is raised inside a GC finalizer and dropped — there is no exception in user code, just a stalled event loop.
- Triggers in real workloads via the `@observe` decorator when a logged payload happens to be cyclic (e.g. an OAuth token-refresh log record on FastAPI workers).

## Changes
- `langfuse/_utils/serializer.py` — id-based cycle guard added to the `dict`, `list`, `Sequence`, and `__slots__` branches (mirroring the existing `__dict__` guard). The `__dict__` branch's `set.remove` was switched to `set.discard` inside a `try` / `finally` so an exception mid-walk no longer leaks `seen` state.
- `tests/unit/test_serializer_cycle_detection.py` — 15 new regression tests covering:
  - direct self-cycles on each branch (`dict`, `list`, custom `Sequence`, `__slots__`-only object)
  - mutual cycles (`a → b → a`) and mixed-container cycles (`dict → list → dict`)
  - shared-object DAGs are **not** flagged as cycles
  - 50-level non-cyclic chains serialise fully
  - back-to-back encodes on the same instance do not leak `seen` state
  - the existing `__dict__` cycle marker (`type(obj).__name__`) is preserved
  - the encode of a cyclic dict completes in under 1 s (the GIL-starvation symptom)

## Test plan
- [x] `uv run --frozen pytest tests/unit/test_serializer_cycle_detection.py tests/unit/test_serializer.py -v` — 33 / 33 pass
- [x] `uv run --frozen pytest -n auto --dist worksteal tests/unit` — 435 pass, 2 skipped, 0 fail
- [x] `uv run --frozen ruff check .` (changed files) — clean
- [x] `uv run --frozen ruff format --check .` (changed files) — already formatted
- [x] `uv run --frozen mypy langfuse --no-error-summary` — clean

This PR fixes a silent production hang (issue #1655) where cyclic object graphs passed through `@observe` could stall the asyncio event loop: the `RecursionError` was swallowed inside Python's GC, never surfacing to user code. The same `id()`-based cycle guard that already existed for the `__dict__` branch is now applied to the `dict`, `list`, `Sequence`, and `__slots__` branches, each wrapped in `try/finally` + `set.discard` so that shared-but-non-cyclic objects (DAGs) are not incorrectly flagged.

- The four previously unprotected branches (`dict`, `list`, `Sequence`, `__slots__`) now detect cycles and return a `\"<cycle:Type>\"` sentinel instead of recursing; the pre-existing `__dict__` marker format (`type(obj).__name__`) is left unchanged for back-compat.
- The `__dict__` branch is hardened: `set.remove` (which can raise `KeyError` if state is corrupt) is replaced with `set.discard` inside a `try/finally`, matching the new branches.
- 15 regression tests cover self-cycles, mutual cycles, mixed-container cycles, DAG correctness, back-to-back encode state isolation, and a timing assertion guarding against the original GIL-starvation symptom.
</details>

Safe to merge. Changes are additive cycle guards on four previously unprotected branches, each using the same try/finally + discard pattern so DAGs are not incorrectly flagged, and the test suite is thorough.

One test comment incorrectly describes which branch is exercised for _CycleSequence — the Sequence branch fires before __dict__, not after — which could mislead future maintainers tracing a regression. The production code itself is correct; this is a documentation gap in the test file only.

The comment at line 203 of tests/unit/test_serializer_cycle_detection.py should be corrected before merge.
</details>

```mermaid
flowchart TD
    A["EventSerializer.default(obj)"] --> B{Type dispatch}
    B -->|tuple/set/frozenset| C["return list(obj)\nno cycle guard needed"]
    B -->|dict| D{"id in seen?"}
    B -->|list| E{"id in seen?"}
    B -->|Sequence| F{"id in seen?"}
    B -->|has slots| G{"id in seen?"}
    B -->|has dict| H{"id in seen?"}
    D -->|yes| I["return cycle:dict sentinel"]
    D -->|no| J["add id to seen\nrecurse keys+values\nfinally discard id"]
    E -->|yes| K["return cycle:list sentinel"]
    E -->|no| L["add id to seen\nrecurse items\nfinally discard id"]
    F -->|yes| M["return cycle:Sequence sentinel"]
    F -->|no| N["add id to seen\nrecurse items\nfinally discard id"]
    G -->|yes| O["return cycle:SlotType sentinel"]
    G -->|no| P["add id to seen\nrecurse slots dict\nfinally discard id"]
    H -->|yes| Q["return type name\nback-compat marker"]
    H -->|no| R["add id to seen\nrecurse __dict__\nfinally discard id"]
```
</details>

`````markdown

---

### Issue 1 of 1
tests/unit/test_serializer_cycle_detection.py:203-204
The comment is incorrect about which branch is exercised. `_CycleSequence` inherits directly from `collections.abc.Sequence`, so `isinstance(seq, Sequence)` is `True` and the `Sequence` branch (line 157 of `serializer.py`) is reached before the `__dict__` branch is ever considered. The `__dict__` branch is only reachable via the `elif`, which is skipped once the earlier `isinstance(obj, Sequence)` guard fires. The test verifies valid behaviour regardless, but the comment misdirects future readers about which code path is under test.

```suggestion
    # _CycleSequence inherits from collections.abc.Sequence, so the
    # Sequence branch in EventSerializer.default is hit first (before __dict__);
    # either way the encode must terminate and produce valid JSON.
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(serializer): detect cycles in dict/l..."](https://github.com/langfuse/langfuse-python/commit/657cc2e883aacf6391fef26380620de77ac01d0f) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32665350)</sub>

### DIFF
--- a/langfuse/_utils/serializer.py
+++ b/langfuse/_utils/serializer.py
@@ -126,21 +126,55 @@ class EventSerializer(JSONEncoder):
             if isinstance(obj, (tuple, set, frozenset)):
                 return list(obj)
 
+            # Cycle detection for the container branches. Without these guards a
+            # cyclic dict/list/Sequence/__slots__ graph recurses until the
+            # RecursionError is swallowed inside Python's GC, leaving the asyncio
+            # loop GIL-starved (issue #1655). try/finally + discard keeps a
+            # shared object reachable from sibling subtrees (a DAG, not a cycle)
+            # from being mis-marked.
             if isinstance(obj, dict):
-                return {self.default(k): self.default(v) for k, v in obj.items()}
+                obj_id = id(obj)
+                if obj_id in self.seen:
+                    return f"<cycle:{type(obj).__name__}>"
+                self.seen.add(obj_id)
+                try:
+                    return {self.default(k): self.default(v) for k, v in obj.items()}
+                finally:
+                    self.seen.discard(obj_id)
 
             if isinstance(obj, list):
-                return [self.default(item) for item in obj]
+                obj_id = id(obj)
+                if obj_id in self.seen:
+                    return f"<cycle:{type(obj).__name__}>"
+                self.seen.add(obj_id)
+                try:
+                    return [self.default(item) for item in obj]
+                finally:
+                    self.seen.discard(obj_id)
 
             # Important: this needs to be always checked after str and bytes types
             # Useful for serializing protobuf messages
             if isinstance(obj, Sequence):
-                return [self.default(item) for item in obj]
+                obj_id = id(obj)
+                if obj_id in self.seen:
+                    return f"<cycle:{type(obj).__name__}>"
+                self.seen.add(obj_id)
+                try:
+                    return [self.default(item) for item in obj]
+                finally:
+                    self.seen.discard(obj_id)
 
             if hasattr(obj, "__slots__"):
-                return self.default(
-                    {slot: getattr(obj, slot, None) for slot in obj.__slots__}
-                )
+                obj_id = id(obj)
+                if obj_id in self.seen:
+                    return f"<cycle:{type(obj).__name__}>"
+                self.seen.add(obj_id)
+                try:
+                    return self.default(
+                        {slot: getattr(obj, slot, None) for slot in obj.__slots__}
+                    )
+                finally:
+                    self.seen.discard(obj_id)
             elif hasattr(obj, "__dict__"):
                 obj_id = id(obj)
 
@@ -149,8 +183,10 @@ class EventSerializer(JSONEncoder):
                     return type(obj).__name__
                 else:
                     self.seen.add(obj_id)
-                    result = {k: self.default(v) for k, v in vars(obj).items()}
-                    self.seen.remove(obj_id)
+                    try:
+                        result = {k: self.default(v) for k, v in vars(obj).items()}
+                    finally:
+                        self.seen.discard(obj_id)
 
                     return result
 
--- a/tests/unit/test_serializer_cycle_detection.py
+++ b/tests/unit/test_serializer_cycle_detection.py
@@ -1,0 +1,215 @@
+"""Regression tests for issue #1655 — cycle detection on the dict / list /
+Sequence / __slots__ branches of EventSerializer. Pre-fix these branches
+recursed forever, the RecursionError was swallowed inside GC, and the
+asyncio loop was GIL-starved for minutes."""
+
+import json
+import time
+from collections.abc import Sequence
+
+from langfuse._utils.serializer import EventSerializer
+
+
+def test_dict_self_cycle_does_not_recurse():
+    d = {"a": 1}
+    d["self"] = d
+
+    parsed = json.loads(EventSerializer().encode(d))
+
+    assert parsed["a"] == 1
+    assert "cycle" in parsed["self"]
+    assert "dict" in parsed["self"]
+
+
+def test_list_self_cycle_does_not_recurse():
+    lst = [1, 2, 3]
+    lst.append(lst)
+
+    parsed = json.loads(EventSerializer().encode(lst))
+
+    assert parsed[:3] == [1, 2, 3]
+    assert "cycle" in parsed[3]
+    assert "list" in parsed[3]
+
+
+def test_slots_self_cycle_does_not_recurse():
+    """__slots__-only object (no __dict__) with a self-referential slot."""
+
+    class SlotsObj:
+        __slots__ = ("name", "ref")
+
+        def __init__(self):
+            self.name = "outer"
+            self.ref = None
+
+    obj = SlotsObj()
+    obj.ref = obj
+
+    parsed = json.loads(EventSerializer().encode(obj))
+
+    assert parsed["name"] == "outer"
+    assert "cycle" in parsed["ref"]
+
+
+def test_cycle_encode_completes_quickly():
+    """Pre-fix this hung the interpreter at the recursion limit while the
+    asyncio loop sat GIL-starved. With the guard it returns instantly."""
+    d = {"a": 1}
+    d["self"] = d
+
+    start = time.monotonic()
+    EventSerializer().encode(d)
+    elapsed = time.monotonic() - start
+
+    assert elapsed < 1.0, f"encode took {elapsed:.3f}s — cycle detection regressed"
+
+
+def test_mutual_dict_cycle():
+    """a -> b -> a — cycle straddles two distinct dict ids."""
+    a = {}
+    b = {}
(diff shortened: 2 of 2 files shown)


hassiebp:
Thanks for raising this, this should be already fixed on main

PR #1658, opened 2026-05-19
fix(resource_manager): reinitialize consumer threads after os.fork()

pyg410:
## What does this PR do?

When using Gunicorn with `--preload`, `os.fork()` copies memory but not threads
(POSIX.1: https://pubs.opengroup.org/onlinepubs/9699919799/functions/fork.html).
`LangfuseResourceManager` starts media upload and score ingestion consumer threads
on initialization, but these threads are not inherited by forked worker processes.
As a result, all media upload and score ingestion events are silently lost, and
calling `flush()` blocks forever on `queue.join()` → Gunicorn worker timeout (SIGABRT).

Note: span export is already fork-safe via `LangfuseSpanProcessor` (`BatchSpanProcessor`
inheritance). This fix covers the remaining background threads managed by
`LangfuseResourceManager`.

Related PR
- #1654 (initial fork-safety fix in v2-stable)

**Changes:**
- Extract consumer thread initialization into `_init_consumer_threads()` for reuse
- Store `score_ingestion_client` as an instance variable for access in `_at_fork_reinit()`
- Add `_at_fork_reinit()` to reinitialize queues and consumer threads after fork
- Register `os.register_at_fork(after_in_child=...)` using `weakref.WeakMethod` to
  avoid permanent strong references that would prevent garbage collection
- Add `_shutdown` flag to skip reinitialization on already-stopped instances

## Type of change

- [x] Bug fix

## Verification

List the main commands you ran:

```bash
 uv run pytest tests/unit/test_resource_manager.py -v
```

## Checklist

This PR fixes silent data loss and `flush()` deadlocks in Gunicorn `--preload` deployments by registering an `os.register_at_fork(after_in_child=...)` handler that reinitializes the media upload and score ingestion consumer threads in each forked worker.

- Consumer thread startup is extracted into `_init_consumer_threads()` so it can be called both during initial setup and after fork; queues are intentionally recreated in the child to prevent duplicating pre-fork events across workers.
- A `WeakMethod`-based fork handler prevents the `os.register_at_fork` registry from holding a strong reference that would block garbage collection, and a `_shutdown` flag guards against reinitializing already-torn-down instances.
- The fix deliberately omits recreating `httpx.Client` / `LangfuseClient` instances after fork (acknowledged in comments), leaving inherited connection-pool FDs that could cause initial connection errors in workers; the `after_in_child` callback also has no exception handling, so a thread-creation failure propagates out of `os.fork()` and crashes the worker rather than degrading gracefully.
</details>

The core fix is sound and addresses a real production issue, but the `after_in_child` callback has no exception handling, meaning a thread-creation failure would crash the Gunicorn worker rather than falling back gracefully.

The fork-reinitialization logic is correct for the happy path: queues are replaced, old dead thread references are discarded, new consumer threads are started, and the WeakMethod approach is appropriate. The main gap is that `_init_consumer_threads()` is called directly inside the `after_in_child` callback without a try/except — if the OS refuses to create a thread at worker startup time, the exception surfaces from `os.fork()` and kills the worker instead of allowing it to continue with degraded telemetry. The inherited httpx connection pool FDs are an acknowledged limitation that could surface as transient connection errors in workers.

`langfuse/_client/resource_manager.py` warrants a second look around the `_at_fork_reinit` method — specifically exception handling and whether the HTTP clients need to be recreated for full fork safety.
</details>

```mermaid
sequenceDiagram
    participant Master as Master Process (Gunicorn)
    participant Fork as os.fork()
    participant Parent as Parent Process
    participant Child as Child Worker Process

    Master->>Master: Langfuse() → LangfuseResourceManager.__new__()
    Master->>Master: _initialize_instance()
    Master->>Master: _init_consumer_threads() → start media + score threads
    Master->>Master: "os.register_at_fork(after_in_child=weak_reinit_lambda)"
    Master->>Fork: os.fork()
    Fork-->>Parent: returns child PID (old threads alive)
    Fork-->>Child: returns 0 (threads NOT inherited)
    Child->>Child: after_in_child: weak_reinit() → _at_fork_reinit()
    Child->>Child: _shutdown? → False → proceed
    Child->>Child: _init_consumer_threads()
    Child->>Child: new _media_upload_queue (fresh Queue)
    Child->>Child: new _score_ingestion_queue (fresh Queue)
    Child->>Child: start fresh MediaUploadConsumer threads
    Child->>Child: start fresh ScoreIngestionConsumer thread
    Child->>Child: child ready to handle requests
    Parent->>Parent: continues with original threads unchanged
```
</details>

`````markdown

---

### Issue 1 of 3
langfuse/_client/resource_manager.py:344-378
**Unhandled exception in `after_in_child` callback crashes Gunicorn worker**

`_at_fork_reinit` calls `_init_consumer_threads`, which calls `Thread.start()`. If the OS refuses to create a thread — e.g., due to resource exhaustion (`OSError: can't start new thread`) — the exception propagates through the `after_in_child` callback chain and surfaces as an exception from `os.fork()` in the child process. Gunicorn would see the worker fail immediately at startup rather than the telemetry path degrading gracefully. Wrapping `_init_consumer_threads()` in a `try/except Exception` and logging the error would allow the child to continue (without consumer threads) instead of crashing.

### Issue 2 of 3
langfuse/_client/resource_manager.py:366-374
**Inherited `httpx.Client` connection pool is not fork-safe**

`self.httpx_client` and `self._score_ingestion_client` (which holds a reference to `self.httpx_client` as its `session`) are not recreated in `_at_fork_reinit`. After `os.fork()`, the child inherits duplicated file descriptors from the parent's connection pool. The parent and child share open TCP connections at the OS level; if the parent closes or recycles a connection, the child's pool may try to reuse a dead FD, causing connection errors on the first outbound requests from each worker. The existing comment acknowledges this but marks it as a future concern — since this PR is specifically targeting fork-safety, recreating the clients in the child would complete the fix.

### Issue 3 of 3
langfuse/_client/resource_manager.py:80-82
**Class-level `_lock` (RLock) not guarded around fork**

`LangfuseResourceManager._lock` is a `threading.RLock`. If `os.fork()` occurs while another thread holds `_lock` (i.e., during a concurrent `__new__` call), the child inherits the locked mutex with no thread to release it. Any subsequent `LangfuseResourceManager.__new__` call in the child — including indirect calls — would deadlock. The standard remedy is to register a `before` handler (`os.register_at_fork(before=cls._lock.acquire)`) and release it in both `after_in_parent` and `after_in_child`. In a typical Gunicorn `--preload` deployment this lock is already released before workers fork, so this is a latent risk rather than an immediate failure for the primary use case.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(resource\_manager): reinitialize cons..."](https://github.com/langfuse/langfuse-python/commit/590963318edd99d4e75d349aac2bdce521cb1eb3) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=32819925)</sub>

### CODE REVIEW

File: langfuse/_client/resource_manager.py [hunk 1]
```diff
@@ -266,48 +316,71 @@ def _initialize_instance(
...
+            flush_interval=self.flush_interval,
             max_retries=3,
             public_key=self.public_key,
         )
         ingestion_consumer.start()
         self._ingestion_consumers.append(ingestion_consumer)
 
-        # Register shutdown handler
-        atexit.register(self.shutdown)
+    def _at_fork_reinit(self) -> None:
```
wochinge, line 402: I think we also need re-init `_lock = threading.RLock()` because it might be locked when the forking is happening and then it can never be released in the forked child as there is no owning thread in that child.

langfuse/_client/resource_manager.py
```diff
@@ -266,48 +316,71 @@ def _initialize_instance(
...
+
+        langfuse_logger.debug(
+            f"[PID {os.getpid()}] Fork detected: reinitializing Langfuse consumer threads."
+        )
+
+        # Queues are intentionally recreated after fork. Items enqueued before fork
+        # belong to the preloaded parent process and must not be processed by every
+        # worker — otherwise uploads/scores would be duplicated across workers.
+        #
+        # HTTP clients (self.httpx_client, self._score_ingestion_client) are not recreated
```
wochinge: I'm afraid we need to recreate them too.
From what I saw in the docs for `httpx` it is promising thread-safety (which makes the singleton usage fine) but no process-safety. So copying that stuff cross processes might lead to obscure issues.

wochinge (commented):
Great contribution @pyg410 ! 
A few smaller comments - otherwise looking very good already 🙌🏻

langfuse/_client/resource_manager.py
pyg410: > I'm afraid we need to recreate them too. From what I saw in the docs for `httpx` it is promising thread-safety (which makes the singleton usage fine) but no process-safety. So copying that stuff cross processes might lead to obscure issues.

Thanks for the feedback! 
You're right that httpx.Client is not process-safe across fork(). 
We've updated _at_fork_reinit to always recreate self.httpx_client, self.api, and self._score_ingestion_client with a fresh client after fork. 
For the case where a user passes a custom httpx_client: we considered preserving it, but users have no practical way to handle this themselves — there's no public API to replace the client inside the singleton after fork. So after fork we always create a new default client, which means any custom transport/SSL/proxy settings on a user-provided client will not carry over into child processes. 
We're treating this as a known limitation for now: custom httpx_client + fork() (e.g. Gunicorn --preload) is not fully supported. 
A proper fix would require accepting a factory callable (e.g. httpx_client_factory: Callable[[], httpx.Client]) instead of an instance, so the library can recreate the client with the original settings after fork — but this is a public API change that requires broader team discussion before we can proceed. We can follow up on that as a separate improvement. (eba27cc)

wochinge: @pyg410 Thanks for addressing the comments.

Regarding the httpx.Client:
I think I'd do explicitly the opposite: don't reinitialize custom httpx client upon fork (but rather just use the copy).

My reasoning:
Users with custom httpx client can't use the SDK with forking in the current implementation. If we let them pass it (but don't reinitalize) it, then we give them at least the chance to handle the reinitialization themselves (or make the client otherwise process safe)

pyg410: @wochinge 

Thanks! Made the following changes.([157122e](https://github.com/langfuse/langfuse-python/pull/1658/commits/157122e0a346e24692f4aa071b45b16301fd3f94))

- Custom httpx_client is no longer recreated after fork. The fork-inherited copy is reused as-is, so callers retain the opportunity to handle process-safety themselves (e.g. via their own os.register_at_fork(after_in_child=...) handler).
- Stored the original httpx_client reference as self._custom_httpx_client instead of a boolean flag, to avoid potential state drift.
- Updated the docstring to reflect the new behavior.

wochinge:
@pyg410 One more thing: I tried gunicon with `preload` on my Mac and I get a segfault.
Apparently we're doing too heavy stuff in the `after_in_child` - most importantly the http client reinitialization. 

Also had a look at how others do it and seems they all lazy reinitalize the connection.
So I think we need to the same - otherwise forked-usage remains broken

pyg410:
> @pyg410 One more thing: I tried gunicon with `preload` on my Mac and I get a segfault. Apparently we're doing too heavy stuff in the `after_in_child` - most importantly the http client reinitialization.
> 
> Also had a look at how others do it and seems they all lazy reinitalize the connection. So I think we need to the same - otherwise forked-usage remains broken

@wochinge 
Agreed on lazy reinitialization - fixed in the latest commit.(68327b7)
`_at_fork_reinit()` now only does lightweight work safe to call in `after_in_child`(replaces the class lock and swaps in fresh empty queues.)
All heavy work (httpx.Client creation, thread spawning) is deferred to `_ensure_post_fork_initialized()`, which is called lazily on the first `add_score_task()` / `add_trace_task()` in the child process.

wochinge:
@pyg410 Thanks! Just tried it and still breaks. 
The issue is actually not the heavy work but the implementation of `getproxies` on MacOS: https://docs.python.org/3/library/urllib.request.html

I checked a bit how others do it and the simplest options seems to be:
1. Revert last commit (we actually don't need it, and it just makes things more complex)
2. In `_at_fork_reinit`, add the following guard:
   ```python
     if sys.platform == "darwin":
      # urllib proxy discovery uses macOS system APIs that are unsafe after fork.
      # Limit Python's documented workaround to the child process.
      os.environ["no_proxy"] = "*"
      os.environ["NO_PROXY"] = "*"
   ```
   
 This deactivates the proxy discovery on Mac and makes the PR full functional.
 
 Other libraries issues on this:
 - https://github.com/benoitc/gunicorn/issues/2761
 - ddtrace-py has their own http client

wochinge:
@claude review

wochinge:
@claude review

wochinge:
Thanks for the fix @pyg410 ! 🚀 Will do a release with it now.

PR #1688, opened 2026-06-07
fix(observe): preserve current span observation type

### LINKED ISSUE #1676: @observe(as_type=…, capture_output=False) silently downgrades observation to type=span
## Summary

`@observe` accepts both `as_type=…` and `capture_output=False`. Either flag
works correctly in isolation. **Combining them silently records the
observation as `type=span`** instead of the requested type — no warning,
no error, no exception.

In our codebase we needed `as_type="guardrail"` (for Langfuse's typed UI)
and `capture_output=False` (because the body sets a richer output via
`update_current_span(output=…)` and the function's bool return value
would otherwise clobber it). The two flags address orthogonal concerns
and intuitively compose — but using them together silently downgrades
the observation kind, breaking the typed-observation contract.

Result: traces the application author intended to surface in Langfuse's
typed UI (guardrail badges, evaluator dashboards, tool-call filters)
fall back to opaque `span`-typed rows. The intent is lost without any
diagnostic.

## Reproduction (langfuse 4.7.0)

```python
import asyncio

from langfuse import Langfuse, get_client, observe

Langfuse()  # init singleton, keys via env
lf = get_client()


@observe(name="probe.with_capture_output_false", as_type="guardrail", capture_output=False)
async def broken():
    # capture_output=False is needed when the function body sets a richer
    # output via update_current_span(output=...) and the return value would
    # otherwise clobber it.
    lf.update_current_span(output={"verdict": "manually set"})


@observe(name="probe.without_capture_output_false", as_type="guardrail")
async def works():
    return {"verdict": "auto captured"}


async def main():
    with lf.start_as_current_observation(name="probe.root", as_type="span"):
        await broken()
        await works()
    lf.flush()


asyncio.run(main())
```

Observation types after ingestion (verified via `/api/public/observations`):

| Function | Expected | Actual |
|---|---|---|
| `probe.with_capture_output_false` | `GUARDRAIL` | `SPAN` ❌ |
| `probe.without_capture_output_false` | `GUARDRAIL` | `GUARDRAIL` ✓ |

Re-verified on 4.7.0 (released 2026-05-27). Also reproduces on 4.6.1 and 4.0.6.

## Why this is a problem

1. The two flags are orthogonal in intent. `as_type` selects the
   UI/analytics treatment (typed icon, filter chip, dashboard
   aggregation). `capture_output=False` opts out of auto-output-capture
   so an explicit `update_current_span(output=…)` from inside the
   function body isn't clobbered. They naturally compose for any
   guardrail / evaluator / tool wrapper that wants a typed observation
   AND a custom output shape.
2. Typed observations gate a meaningful chunk of Langfuse-native UX —
   guardrail status badges, evaluator score aggregations in the
   Datasets/Evals view, per-type filter chips in the trace list,
   tool-call rendering. A silent downgrade means none of those views
   pick up the observation.
3. The downgrade is silent. No warning, no log line, no exception.
   Trivially missed in code review and in deployed code; you only
   notice when the observation doesn't appear under the expected
   filter in the Langfuse UI.
4. The workaround (drop `@observe`, use explicit
   `start_as_current_observation`) is more verbose and loses the
   decorator's signature-preservation and async-vs-sync auto-handling.
   Doable, but the decorator should support the same composition.
5. The combination isn't documented as unsupported in the SDK
   reference. It's a runtime contradiction, not a documented
   constraint.

## Suggested fix

Either of:

1. **Honour `as_type` regardless of `capture_output`** (preferred).
   The OTel observation kind and the output-capture mechanism aren't
   coupled in the underlying span model — setting one shouldn't affect
   the other. Likely a small change in the decorator path where the
   typed-as kind is dropped when `capture_output=False` causes the
   span builder to take a different branch.
2. **Raise at decoration time** if the combination is genuinely
   unsupported, with a message pointing to the explicit-observation
   workaround. Beats silent fallback.

A silent type downgrade is the worst of the three options.

## Version info

- langfuse-python: **4.7.0** (also reproduces on 4.6.1 and 4.0.6)
- Python: 3.11.12
- macOS 14, `cloud.langfuse.com` backend

## Current workaround we use

Replace the `@observe` decorator with explicit
`start_as_current_observation`, and call `obs.update(output=…)` inside
the with-block:

```python
async def run_output_guardrails(...):
    lf = get_client()
    with lf.start_as_current_observation(
        name="output_guardrails",
        as_type="guardrail",
        input={...},
    ) as obs:
        result = await guardrails_service.check_message(...)
        obs.update(output={
            "blocked": not result.safe,
            "safe": result.safe,
            "violations": result.violations,
            "checkers": [...],
        })
        return not result.safe
```

This works — both `as_type` and the explicit output stick. It's
strictly more boilerplate than the `@observe` form would be, and you
give up the decorator's signature preservation.

Happy to send a PR — the cleanest path is probably preserving the
typed-as kind on the OTel observation regardless of `capture_output`,
since the two are not coupled in the underlying span model.

fengjikui:
## Summary

- Preserve the active observation type when `update_current_span()` wraps the current OTEL span for updates.
- Add a regression test for `@observe(as_type="guardrail", capture_output=False)` followed by `update_current_span(output=...)`.

Fixes #1676.

## Verification

- `uv run --frozen pytest tests/unit/test_observe.py::test_observe_capture_output_false_preserves_as_type_after_current_span_update -q`
- `uv run --frozen pytest tests/unit/test_observe.py -q`
- `uv run --frozen pytest tests/unit/test_otel.py::TestBasicSpans::test_update_current_span_name tests/unit/test_otel.py::TestBasicSpans::test_update_current_generation_name -q`
- `uv run --frozen ruff check langfuse/_client/client.py tests/unit/test_observe.py`
- `uv run --frozen ruff format --check langfuse/_client/client.py tests/unit/test_observe.py`

This PR fixes a bug where `update_current_span()` always wrapped the active OTEL span as a `LangfuseSpan`, losing the original observation type (e.g., `"guardrail"`) set by `@observe(as_type=...)`. The fix reads the `OBSERVATION_TYPE` attribute from the current OTEL span and dispatches to `_get_span_class()` before constructing the wrapper — consistent with the pattern already used in `set_current_trace_io` and `set_current_trace_as_public`.

- **`langfuse/_client/client.py`**: `update_current_span` now reads `LangfuseOtelSpanAttributes.OBSERVATION_TYPE` from the active span's attributes and uses `_get_span_class()` to select the correct wrapper class instead of hardcoding `LangfuseSpan`.
- **`tests/unit/test_observe.py`**: Adds a regression test verifying that a `@observe(as_type=\"guardrail\", capture_output=False)` span retains its type after `update_current_span(output=...)` is called.
</details>

Safe to merge — the change is a targeted, one-line logical fix that matches the existing pattern used by two other methods in the same class.

The fix reads the existing OBSERVATION_TYPE attribute before constructing the wrapper span, exactly mirroring set_current_trace_io and set_current_trace_as_public which already did this correctly. The _get_span_class method has a safe fallback to LangfuseSpan for unknown types. The added regression test directly covers the reported bug scenario.

No files require special attention.
</details>

```mermaid
sequenceDiagram
    participant U as User Code
    participant D as @observe(as_type=guardrail)
    participant C as Langfuse Client
    participant OS as OTEL Span

    U->>D: call decorated function
    D->>OS: "start span, set OBSERVATION_TYPE=guardrail"
    U->>C: "update_current_span(output=...)"
    C->>OS: _get_current_otel_span()
    OS-->>C: current_otel_span
    C->>OS: attributes.get(OBSERVATION_TYPE, span)
    OS-->>C: guardrail
    C->>C: _get_span_class(guardrail) to LangfuseGuardrail
    C->>C: "LangfuseGuardrail(otel_span=current_otel_span)"
    C->>OS: "span.update(output=...)"
    Note over OS: OBSERVATION_TYPE preserved as guardrail
    D->>OS: "end span (capture_output=False)"
```
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(observe): preserve current span obse..."](https://github.com/langfuse/langfuse-python/commit/0d44ccf449a46d8e5ffd01b5e0b2a6608a8894cb) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36085131)</sub>

### DIFF
--- a/langfuse/_client/client.py
+++ b/langfuse/_client/client.py
@@ -1378,7 +1378,13 @@ class Langfuse:
         current_otel_span = self._get_current_otel_span()
 
         if current_otel_span is not None:
-            span = LangfuseSpan(
+            existing_observation_type = current_otel_span.attributes.get(  # type: ignore[attr-defined]
+                LangfuseOtelSpanAttributes.OBSERVATION_TYPE, "span"
+            )
+            span_class = self._get_span_class(
+                cast(ObservationTypeLiteral, existing_observation_type)
+            )
+            span = span_class(
                 otel_span=current_otel_span,
                 langfuse_client=self,
                 environment=self._environment,
--- a/tests/unit/test_observe.py
+++ b/tests/unit/test_observe.py
@@ -114,6 +114,34 @@ async def test_streaming_response_preserves_context_without_output_capture(
     assert LangfuseOtelSpanAttributes.OBSERVATION_OUTPUT not in endpoint_span.attributes
 
 
+def test_observe_capture_output_false_preserves_as_type_after_current_span_update(
+    langfuse_memory_client: Any, memory_exporter: Any
+) -> None:
+    @observe(
+        name="probe.with_capture_output_false",
+        as_type="guardrail",
+        capture_output=False,
+    )
+    def probe() -> dict[str, bool]:
+        langfuse_memory_client.update_current_span(output={"verdict": "manually set"})
+
+        return {"decorator_output": True}
+
+    assert probe() == {"decorator_output": True}
+
+    langfuse_memory_client.flush()
+
+    probe_span = _finished_spans_by_name(
+        memory_exporter, "probe.with_capture_output_false"
+    )[0]
+
+    assert (
+        probe_span.attributes[LangfuseOtelSpanAttributes.OBSERVATION_TYPE]
+        == "guardrail"
+    )
+    assert LangfuseOtelSpanAttributes.OBSERVATION_OUTPUT in probe_span.attributes
+
+
 def test_sync_generator_wrapper_close_ends_span_without_exhaustion() -> None:
     def generator() -> Generator[str, None, None]:
         yield "item_0"


fengjikui:
Closing this because #1705 was merged and explicitly fixed #1676 with the same observation-type preservation path. This PR is now conflicting and superseded, so keeping it open would add duplicate review noise. Thanks for confirming the issue and landing the fix.

PR #1689, opened 2026-06-07
fix(langchain): silence intentional placeholder warnings

### LINKED ISSUE #1667: Suppress "Placeholders have not been resolved" warning when MessagesPlaceholder use is intentional (LangChain integration)
## Summary

`ChatPromptClient.compile()` (and `get_langchain_prompt()` which calls it
under the hood) unconditionally logs:

> `Placeholders ['...'] have not been resolved. Pass them as keyword arguments to compile().`

even when leaving placeholders unresolved is **intentional** — i.e. the
caller follows the officially documented LangChain integration pattern
where unresolved placeholders are deliberately returned as
`MessagesPlaceholder` for LangChain's `ChatPromptTemplate` to resolve at
`.invoke()` time.

Result: production logs are flooded with one WARNING per request, masking
real signals.

## Reproduction (langfuse 4.6.1, official docs example)

```python
from langfuse import get_client
from langchain_core.prompts import ChatPromptTemplate

langfuse = get_client()

# Create a chat prompt with a placeholder (per docs example)
langfuse.create_prompt(
    name="movie-critic-chat",
    type="chat",
    prompt=[
        {"role": "system", "content": "You are an expert movie critic"},
        {"type": "placeholder", "name": "chat_history"},
        {"role": "user", "content": "What should I watch next?"},
    ],
    labels=["production"],
)

# Documented LangChain integration:
# https://langfuse.com/docs/prompt-management/features/message-placeholders
prompt = langfuse.get_prompt("movie-critic-chat")
langchain_prompt = ChatPromptTemplate.from_template(prompt.get_langchain_prompt())

# → logs WARNING: Placeholders ['chat_history'] have not been resolved.
#   even though leaving it unresolved is EXACTLY what the docs tell you
#   to do for LangChain integration.
```

## Why this is a problem

1. The docs at `/docs/prompt-management/features/message-placeholders`
   (LangChain Python tab) endorse this exact pattern — `MessagesPlaceholder`
   objects in the output are the documented goal.
2. SDK can't distinguish "intentional unresolved (LangChain hand-off)" from
   "misuse (forgot kwargs)" and warns in both cases.
3. No public API to silence: no `silent=` / `suppress_warnings=` kwarg on
   `compile()` or `get_langchain_prompt()`.
4. `logging.getLogger("langfuse").setLevel(...)` is not a reliable
   workaround — see langfuse/langfuse#4976 reporting setLevel doesn't
   affect certain SDK-emitted warnings.
5. Forces every LangChain integrator to install a `logging.Filter` that
   string-matches the warning message — brittle (breaks on copy edits)
   and easy to over-silence.

## Suggested fix

Add a `silent=True` (or `suppress_unresolved_warning=True`) kwarg to:

- `ChatPromptClient.compile(**kwargs, silent: bool = False)`
- `ChatPromptClient.get_langchain_prompt(**kwargs, silent: bool = False)`

When `silent=True`, the
"Placeholders ['...'] have not been resolved" warning is skipped, but
other validation warnings (e.g. *"Placeholder X must contain a list of
chat messages"*) still fire — those are genuine misuse signals.

Alternatively, `get_langchain_prompt()` could pass `silent=True` to its
internal `compile()` call by default, since by definition any unresolved
placeholder returned from this method is intended for LangChain.

## Version info

- langfuse-python: 4.6.1 (also reproduces on 4.5.1 + main `HEAD`)
- langchain-core: 1.3.2
- Python: 3.11

## Current workaround we use

`contextlib.contextmanager` that scopes a `logging.Filter` matching the
substring `"have not been resolved"` over the
`chat_skeleton.get_langchain_prompt()` call site. Fragile (depends on
warning string), but scoped narrowly so other langfuse warnings still
propagate.

Happy to send a PR if the maintainers agree with the kwarg shape.

fengjikui:
## Summary

- Keep unresolved-placeholder warnings for direct `ChatPromptClient.compile()` calls.
- Stop emitting that warning from `get_langchain_prompt()` when unresolved placeholders are intentionally returned as LangChain `MessagesPlaceholder` objects.
- Add regression coverage to ensure the LangChain hand-off path stays quiet while preserving the existing placeholder output.

Fixes #1667.

## Verification

- `uv run --frozen pytest tests/unit/test_prompt_compilation.py::TestLangchainPromptCompilation::test_get_langchain_prompt_with_unresolved_placeholders -q`
- `uv run --frozen pytest tests/unit/test_prompt_compilation.py -q`
- `uv run --frozen ruff check langfuse/model.py tests/unit/test_prompt_compilation.py`
- `uv run --frozen ruff format --check langfuse/model.py tests/unit/test_prompt_compilation.py`
- `git diff --check`

I also tried `uv run --frozen pytest tests/e2e/test_prompt.py::test_warning_on_unresolved_placeholders -q`, but the local run requires Langfuse server credentials/API setup and failed before the assertion path with `Langfuse client initialized without public_key`.

This PR fixes spurious `langfuse_logger.warning` calls that fire during `get_langchain_prompt()` when unresolved placeholders are intentionally left for LangChain to handle as `MessagesPlaceholder` objects.

- `compile()` is refactored into a thin wrapper around a new private `_compile(warn_on_unresolved_placeholders: bool)` method; the public API is unchanged and still warns by default.
- `get_langchain_prompt()` now calls `_compile(warn_on_unresolved_placeholders=False)`, suppressing the warning only on the LangChain hand-off path.
- A regression test asserts that `langfuse_logger.warning` is never called during `get_langchain_prompt()` when a placeholder is intentionally left unresolved.
</details>

Safe to merge; the behavioural change is narrow and well-tested.

The refactor correctly isolates warning behaviour behind a flag, public compile() is unchanged, and a targeted regression test covers the silent path. The only finding is a style-level import placed inside a test method body rather than at the module top.

No files require special attention beyond the minor import placement in the test.
</details>

```mermaid
flowchart TD
    A["compile(kwargs)"] -->|"warn=True"| B["_compile()"]
    C["get_langchain_prompt(kwargs)"] -->|"warn=False"| B

    B --> D{placeholder resolved?}
    D -- Yes --> E[expand messages]
    D -- No --> F[keep as placeholder dict]
    F --> G{warn flag set?}
    G -- Yes --> H["langfuse_logger.warning()"]
    G -- No --> I[silent]

    A --> L[return raw dicts]
    C --> M["convert to LangChain tuples / MessagesPlaceholder"]
```
</details>

`````markdown

---

### Issue 1 of 1
tests/unit/test_prompt_compilation.py:805-810
The `patch` import is placed inside the test method body. The project rule requires imports to be at the top of the module. The two pre-existing local imports (`Prompt_Chat`, `ChatPromptClient`) also violate this rule, but this PR actively adds a third one (`from unittest.mock import patch`) and is a good opportunity to move at least the new import to the top-level block.

```suggestion
    def test_get_langchain_prompt_with_unresolved_placeholders(self):
        """Test that unresolved placeholders become MessagesPlaceholder objects."""
        from langfuse.api import Prompt_Chat
        from langfuse.model import ChatPromptClient
```

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["fix(langchain): silence intentional plac..."](https://github.com/langfuse/langfuse-python/commit/43f649f3519c2db2e6c67c172368bf7defa21f58) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=36088848)</sub>

**Context used:**

- Rule used - Move imports to the top of the module instead of p... ([source](https://app.greptile.com/personal-org-4986/-/custom-context?memory=c960fc07-9928-409f-a18b-a780cbdded12))

**Learned From**

### DIFF
--- a/langfuse/model.py
+++ b/langfuse/model.py
@@ -309,6 +309,18 @@ class ChatPromptClient(BasePromptClient):
             Dict[str, Any], ChatMessageDict, ChatMessageWithPlaceholdersDict_Placeholder
         ]
     ]:
+        return self._compile(warn_on_unresolved_placeholders=True, **kwargs)
+
+    def _compile(
+        self,
+        *,
+        warn_on_unresolved_placeholders: bool,
+        **kwargs: Union[str, Any],
+    ) -> Sequence[
+        Union[
+            Dict[str, Any], ChatMessageDict, ChatMessageWithPlaceholdersDict_Placeholder
+        ]
+    ]:
         """Compile the prompt with placeholders and variables.
 
         Args:
@@ -381,7 +393,7 @@ class ChatPromptClient(BasePromptClient):
                     compiled_messages.append(chat_message)
                     unresolved_placeholders.append(chat_message["name"])  # type: ignore
 
-        if unresolved_placeholders:
+        if warn_on_unresolved_placeholders and unresolved_placeholders:
             unresolved_placeholders_message = f"Placeholders {unresolved_placeholders} have not been resolved. Pass them as keyword arguments to compile()."
             langfuse_logger.warning(unresolved_placeholders_message)
 
@@ -446,7 +458,10 @@ class ChatPromptClient(BasePromptClient):
             List of messages in the format expected by Langchain's ChatPromptTemplate:
             (role, content) tuples for regular messages or MessagesPlaceholder objects for unresolved placeholders.
         """
-        compiled_messages = self.compile(**kwargs)
+        compiled_messages = self._compile(
+            warn_on_unresolved_placeholders=False,
+            **kwargs,
+        )
         langchain_messages: List[Union[Tuple[str, str], Any]] = []
 
         for msg in compiled_messages:
--- a/tests/unit/test_prompt_compilation.py
+++ b/tests/unit/test_prompt_compilation.py
@@ -804,6 +804,8 @@ Configuration:
 
     def test_get_langchain_prompt_with_unresolved_placeholders(self):
         """Test that unresolved placeholders become MessagesPlaceholder objects."""
+        from unittest.mock import patch
+
         from langfuse.api import Prompt_Chat
         from langfuse.model import ChatPromptClient
 
@@ -826,10 +828,13 @@ Configuration:
         )
 
         # Call get_langchain_prompt without resolving placeholder
-        langchain_messages = prompt_client.get_langchain_prompt(
-            role="helpful",
-            task="coding",
-        )
+        with patch("langfuse.logger.langfuse_logger.warning") as mock_warning:
+            langchain_messages = prompt_client.get_langchain_prompt(
+                role="helpful",
+                task="coding",
+            )
+
+        mock_warning.assert_not_called()
 
         # Should have 3 items: system message, MessagesPlaceholder, user message
         assert len(langchain_messages) == 3

### CODE REVIEW

**Comment by Kairick** [comment]
Ran into the same warning and worked through this patch before finding this PR — the approach looks right to me. Two notes from that, both small:

**1. The flag shares a namespace with template variables.**

`_compile(self, *, warn_on_unresolved_placeholders: bool, **kwargs)` means a prompt that happens to contain a `{{warn_on_unresolved_placeholders}}` variable can no longer be compiled: the value binds to the keyword-only parameter instead of landing in `kwargs`, so the variable is left unsubstituted (and a bool reaches a code path expecting a string).

It's an unlikely name, but public `compile(**kwargs)` currently promises that every keyword is prompt data, and this is the one place that promise gets an exception. Passing the mapping explicitly keeps it intact:

```python
def compile(self, **kwargs: Union[str, Any]) -> ...:
    return self._compile(kwargs, warn_on_unresolved_placeholders=True)

def _compile(
    self,
    kwargs: Dict[str, Any],
    *,
    warn_on_unresolved_placeholders: bool,
) -> ...:
```

and at the call site:

```python
compiled_messages = self._compile(kwargs, warn_on_unresolved_placeholders=False)
```

**2. Two regression tests worth adding.**

The current test covers the quiet path. The two behaviours the issue explicitly asks to preserve aren't pinned yet, so a later refactor could silence them without failing anything:

```python
def test_compile_still_warns_on_unresolved_placeholders(self, caplog):
    with caplog.at_level(logging.WARNING, logger="langfuse"):
        prompt_client.compile()
    assert "Placeholders ['chat_history'] have not been resolved" in caplog.text


def test_get_langchain_prompt_still_warns_on_malformed_placeholder(self, caplog):
    with caplog.at_level(logging.WARNING, logger="langfuse"):
        prompt_client.get_langchain_prompt(chat_history="not-a-list")
    assert "must contain a list of chat messages" in caplog.text
```

`caplog` also avoids patching the shared logger object, so it won't mask a second warning firing in the same call.

Happy to hand over the full test bodies if useful.

PR #1710, opened 2026-06-15
feat(datasets): support multimodal dataset items

wochinge:
## Summary

- Adds multimodal dataset items: `LangfuseMedia` in `input`/`expected_output`/`metadata` is uploaded and replaced with a reference string on `create_dataset_item`.
- `get_dataset(...)` resolves media reference strings to `LangfuseMediaReference` objects (with `fetch_bytes`/`fetch_base64`/`fetch_data_uri` helpers).
- Regenerates the dataset-items + media API clients from the base branch.

## Linear

- [LFE-10289](https://linear.app/langfuse/issue/LFE-10289/python-sdk-changes)

## Major Decisions

- No new external dependency — hand-rolled the small JSONPath subset the API emits instead of pulling in `jsonpath-ng`.
- Always resolve references — only callers of the new API receive them, so an opt-in flag added surface for no benefit.

## Review Focus

- `langfuse/_utils/json_path.py` — the hand-rolled parser (tests generated from jsonpath-plus).
- Generated API client changes are hand-synced from the base branch via fern (spec not yet released).

This PR adds multimodal support to Langfuse dataset items, allowing `LangfuseMedia` objects to be embedded in `input`, `expected_output`, and `metadata` fields and uploaded trace-lessly, and adding a `resolve_media_references=True` option to `get_dataset()` that hydrates backend-provided JSONPath references into `LangfuseMediaReference` objects.

- `create_dataset_item` now recursively discovers `LangfuseMedia` values via `$..`this`` (jsonpath-ng), uploads each one synchronously through the updated `_upload_media_sync` (which accepts `trace_id=None`), deduplicates by `media_id` within a single call, and replaces each value with its reference string before sending to the API.
- `get_dataset(resolve_media_references=True)` passes `include_media_references` to the list endpoint, then walks backend-provided JSONPath annotations to replace reference strings with frozen `LangfuseMediaReference` dataclass instances that expose `fetch_bytes()`, `fetch_base64()`, and `fetch_data_uri()` helpers.
- Generated API types (`DatasetItemMediaReference`, `DatasetItemMediaReferenceField`, `DatasetItemMediaReferenceMedia`) and the `GetMediaUploadUrlRequest` schema are updated to make `trace_id` and `field` optional, enabling trace-less uploads.
</details>

Safe to merge; all changed paths work correctly for the expected JSON-like data shapes, and the two main concerns are style/fragility rather than broken behavior.

The media-upload and hydration flows are logically sound and covered by both unit and E2E tests. The two notable findings are: `_replace_json_path_value` discards the return value of `update()` and relies implicitly on in-place mutation (works today, diverges from how `_process_dataset_item_media` handles the same operation); and `fetch_bytes()` makes a blocking HTTP call with no timeout. Neither produces wrong output in the current code paths, but both leave the door open for subtle failures on refactoring or slow network conditions.

`langfuse/_client/client.py` (`_replace_json_path_value`) and `langfuse/media.py` (`LangfuseMediaReference.fetch_bytes`) are the two spots that would benefit from a second look before shipping to customers at scale.
</details>

```mermaid
sequenceDiagram
    participant User
    participant Langfuse SDK
    participant MediaManager
    participant Langfuse API
    participant Object Storage

    Note over User,Object Storage: create_dataset_item with LangfuseMedia

    User->>Langfuse SDK: create_dataset_item(input={"img": LangfuseMedia(...)})
    Langfuse SDK->>Langfuse SDK: _process_dataset_item_media()<br/>jsonpath-ng discovers LangfuseMedia nodes
    Langfuse SDK->>MediaManager: _upload_media_sync(media)
    MediaManager->>Langfuse API: media.get_upload_url(trace_id=None, field=None)
    Langfuse API-->>MediaManager: upload_url + media_id
    MediaManager->>Object Storage: PUT bytes
    Object Storage-->>MediaManager: 200 OK
    MediaManager->>Langfuse API: media.patch(media_id, upload_http_status=200)
    MediaManager-->>Langfuse SDK: done
    Langfuse SDK->>Langfuse API: dataset_items.create(input={"img": "@@@langfuseMedia:..."})
    Langfuse API-->>Langfuse SDK: DatasetItem
    Langfuse SDK-->>User: DatasetItem (reference strings)

    Note over User,Object Storage: get_dataset with resolve_media_references=True

    User->>Langfuse SDK: get_dataset(name, resolve_media_references=True)
    Langfuse SDK->>Langfuse API: dataset_items.list(includeMediaReferences=true)
    Langfuse API-->>Langfuse SDK: items with mediaReferences[] (JSONPaths + signed URLs)
    loop For each item
        Langfuse SDK->>Langfuse SDK: _hydrate_dataset_item_media_references()<br/>jsonpath-ng update per JSONPath to LangfuseMediaReference
    end
    Langfuse SDK-->>User: DatasetClient (LangfuseMediaReference in fields)
    User->>Langfuse SDK: item.input["img"].fetch_bytes()
    Langfuse SDK->>Object Storage: httpx.get(signed_url)
    Object Storage-->>Langfuse SDK: bytes
    Langfuse SDK-->>User: bytes
```
</details>

`````markdown

---

### Issue 1 of 3
langfuse/_client/client.py:3418-3426
`_replace_json_path_value` silently discards the return value of `update()`, relying entirely on in-place mutation of `value`. This works today because API-deserialized data is always a mutable `dict` or `list`, but it diverges from the pattern used in `_process_dataset_item_media` (which correctly captures `data = match.full_path.update(data, ...)`). If `value` were ever an immutable type at a non-root path, the replacement would be silently lost without any log or exception.

```suggestion
        try:
            value = parse_jsonpath(json_path).update(value, replacement)
        except Exception as e:
            langfuse_logger.debug(
                f"Failed to hydrate dataset media reference at JSONPath {json_path}",
                exc_info=e,
            )

        return value
```

### Issue 2 of 3
langfuse/media.py:50-53
`httpx.get(self.url)` is called without a timeout, so it can block indefinitely if the signed URL is slow or the connection hangs. This is especially relevant in hot paths such as evaluation loops that call `fetch_bytes()` on each dataset item's resolved media.

```suggestion
    def fetch_bytes(self, timeout: float = 30.0) -> bytes:
        """Fetch the media content from the signed URL."""
        response = httpx.get(self.url, timeout=timeout)
        response.raise_for_status()
```

### Issue 3 of 3
langfuse/_task_manager/media_manager.py:233-260
**Silent return on incomplete media is undocumented**

`_upload_media_sync` silently returns `None` when any content field is `None`. The caller `_upload_dataset_item_media` then adds `media_id` to `uploaded_media_ids` and returns the reference string as if the upload succeeded — meaning the dataset item would store a reference to media that was never uploaded.

In practice this path is currently unreachable: the earlier check `if reference_string is None or media_id is None: raise ValueError(...)` in `_upload_dataset_item_media` ensures all content fields are populated before this method is called (because `_media_id` is derived from `_content_sha256_hash`, which requires `_content_bytes`). But the silent return offers no signal to future callers that skip that guard, and no explanation is left in a comment. A guarded `raise` or at minimum a comment explaining the invariant would make this safer.

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["feat(datasets): support media references"](https://github.com/langfuse/langfuse-python/commit/cd791b64eee1677fdddca4070cdd289f4958969a) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=37649631)</sub>

> Greptile also left **3 inline comments** on this PR.

### CODE REVIEW

File: langfuse/_client/client.py
Comment (wochinge) [inline langfuse/_client/client.py:1]: actual (non generated changes)

File: langfuse/_client/client.py [hunk 1]
```diff
@@ -3324,6 +3333,122 @@ def create_dataset_item(
...
+        if self._resources is None:
+            return data
+
+        max_levels = 10
+
+        def _process_data_recursively(
+            data: Any, level: int, ancestor_container_ids: set[int]
+        ) -> Any:
+            # Avoid jsonpath-ng here: dataset writes should keep working
+            # under python -OO where parser docstrings may be stripped.
```
wochinge: in my opinion an okay compromise as the write path doesn't become a hard blocker for users and easy to maintain

Avoid hand rolling jsonpath parsing for the read path -> users can also opt out here so in my opinion fine

langfuse/_client/resource_manager.py
wochinge, line 1: actual (non generated changes)

langfuse/_task_manager/media_manager.py
wochinge, line 1: actual (non generated changes)

langfuse/media.py
wochinge, line 1: actual (non generated changes)

pyproject.toml
```diff
@@ -16,6 +16,7 @@ dependencies = [
     "opentelemetry-api>=1.33.1,<2",
     "opentelemetry-sdk>=1.33.1,<2",
     "opentelemetry-exporter-otlp-proto-http>=1.33.1,<2",
+    "jsonpath-ng>=1.8.0,<2",
```
wochinge: Apache 2.0 license, 731 stars, actively maintained: https://github.com/h2non/jsonpath-ng
Annoying thing is the issues about running optimized python but I think we addressed them enough: https://github.com/h2non/jsonpath-ng#special-note-about-ply-and-docstrings

Next runner up would be `jsonpath-python` (https://pypi.org/project/python-jsonpath/) 
- MIT license 
- only 69 stars

I'd avoid hand rolling the parsing. Especially since we have find references via the TS jsonpath library we don't know exactly which syntax might get in there and which cases of the spec we need to handle.

langfuse/_client/client.py
```diff
@@ -2322,15 +2326,17 @@ def get_dataset(
         *,
         fetch_items_page_size: Optional[int] = 50,
         version: Optional[datetime] = None,
+        resolve_media_references: bool = False,
```
hassiebp: Didn't we decide to drop this option, always resolve and stringify the Media references to the reference string if `str(media)` is called? This would also not be breaking as we would do this only for _new_ dataset items with actual media attached, not legacy ones with mere reference strings

langfuse/_client/client.py
```diff
@@ -2348,8 +2354,16 @@ def get_dataset(
                     page=page,
                     limit=fetch_items_page_size,
                     version=version,
+                    include_media_references=resolve_media_references or None,
```
hassiebp: this option is outdated, no?

langfuse/_client/client.py
wochinge: Yeah, fair. Let me quickly adapt.

langfuse/media.py
```diff
@@ -18,6 +20,81 @@
...
+    """Resolved reference to media stored in Langfuse."""
+
+    media_id: str
+    content_type: str
+    url: str
+    url_expiry: Optional[str] = None
+    content_length: Optional[int] = None
+    reference_string: Optional[str] = None
+
+    def url_is_expired(self) -> bool:
```
hassiebp: nit: is_url_expired might be better naming. also could be a getter

langfuse/media.py
```diff
@@ -18,6 +20,81 @@
...
+        Args:
+            timeout: Request timeout in seconds.
+            client: Optional httpx client to use for the request. Pass this to
+                honor custom transport settings (proxy, CA bundle, mTLS) — in
+                particular when multiple Langfuse clients are configured, since
+                the SDK cannot otherwise tell which client produced this
+                reference. When omitted, the single configured client is used,
+                falling back to a default httpx client.
+        """
+        from langfuse._client.resource_manager import LangfuseResourceManager
```
hassiebp, line 65: why do we have inline import and not top of file?

langfuse/media.py
wochinge: What do you mean by `getter`? A @property or `get_is_url_expired`?

pyproject.toml
hassiebp: Can we get away with simple handrolling? This should not be too complex. For dataset items, we don't allow arbitrary JSON path, but only direct attribute or element access in list. What do you think?

I want to keep dependencies as limited as possible in SDKs and not add one for a peripheral feature

wochinge: I tried but then it quickly became relatively unwieldy. But if keeping the dependency set small is a priority, let me give it another try

hassiebp (changes requested):
Thanks Tobi - only major questions is around the added dependency

langfuse/media.py
wochinge: to avoid circular imports. 

`langfuse.media -> langfuse._client.resource_manager -> langfuse._task_manager.media_manager -> langfuse.media`

langfuse/_client/client.py
wochinge: done

langfuse/_client/client.py
wochinge: done

langfuse/media.py
wochinge: renamed

hassiebp (approved):

hassiebp (approved):

hassiebp (approved):
LGTM

PR #1781, opened 2026-07-26
fix(openai): stop dropping cost when usage is a dict or cost is an int

arthi-arumugam-git:
## What

`_parse_cost` in `langfuse/openai.py` discards cost in two shapes that occur in practice, so `cost_details` comes back `None` and the invocation is recorded with no cost.

```python
if hasattr(usage, "cost") and isinstance(getattr(usage, "cost"), float):
```

**1. `usage` as a dict.** `hasattr({"cost": 0.002}, "cost")` is `False`, so cost is dropped entirely. This is not hypothetical: `_parse_usage` right above it already branches on `isinstance(usage, dict)`, and `_update_langfuse_generation` calls both with the same object:

```python
if usage is not None:
 update["usage_details"] = _parse_usage(usage)
 update["cost_details"] = _parse_cost(usage)
```

So any dict-shaped usage that `_parse_usage` handles correctly loses its cost.

**2. An integer cost.** `isinstance(0, float)` is `False` in Python. OpenRouter reports `"cost": 0` for free models and for fully cached responses, and any whole-number cost serialises as an int. Those get dropped too.

Measured against the current code, only one of five realistic shapes survives:

| input | before | after |
|---|---|---|
| `Usage(cost=0.0021)` | `{"total": 0.0021}` | `{"total": 0.0021}` |
| `Usage(cost=0)` | `None` | `{"total": 0.0}` |
| `Usage(cost=2)` | `None` | `{"total": 2.0}` |
| `{"cost": 0.0021}` | `None` | `{"total": 0.0021}` |
| `{"cost": 0}` | `None` | `{"total": 0.0}` |

## Change

Read `cost` off either shape, and accept `int` as well as `float`. `bool` is excluded explicitly because it is a subclass of `int`, so a stray `"cost": true` would otherwise be recorded as a cost of `1.0`.

## Tests

`tests/unit/test_parse_cost.py`, 8 cases. Two of them fail on the current code and pass with the change:

```
FAILED tests/unit/test_parse_cost.py::test_dict_with_float_cost
FAILED tests/unit/test_parse_cost.py::test_integer_cost_is_not_dropped
2 failed, 6 passed
```

The other six cover `None` usage, an object without `cost`, a dict without `cost`, and the bool and string rejections, so the fix cannot widen the accepted set by accident.

`ruff format --check` is clean on both files. The pre-existing `ruff check` findings in `openai.py` are untouched.

This PR fixes OpenRouter cost extraction and expands regression coverage.
- Reads cost from both dictionary-shaped and object-shaped usage payloads.
- Accepts integer and floating-point costs while explicitly rejecting booleans.
- Normalizes accepted costs to floats and adds tests for supported and rejected values.
</details>

The PR appears safe to merge with no actionable defects identified.

The parser now handles the documented usage shapes while preserving rejection of unsupported values, and the downstream cost-details contract accepts the normalized floating-point result.
</details>

<sub>Reviews (1): Last reviewed commit: ["fix(openai): stop dropping cost when usa..."](https://github.com/langfuse/langfuse-python/commit/532b4b01adef72262af622797c428f5119ea26a1) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=47347043)</sub>

**Context used:**

- Knowledge Base: [OpenAI Integration (langfuse/openai.py)](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse-python/-/docs/openai-integration.md)


arthi-arumugam-git:
Gentle ping: the CLA is signed and the branch is current as of yesterday. Would a maintainer be able to trigger the review bot on this one? Anything needed from my side to move it forward?

PR #1791, opened 2026-07-30
fix(observe): close abandoned generators when their span already ended

DavidTraina:
## What does this PR do?

Once a wrapped generator's span has ended, `aclose()`/`close()` return without closing the generator. But a span can end while the generator is still suspended mid-stream — for example, a client disconnect that cancels iteration before the generator resumes. The generator is then never closed: the GC finalizes it later, running its `finally` block under whatever request context happens to be active at that moment.

In our production streaming service this fired many times a day: "Setting attribute on ended span" warnings from `finally`-block span updates, running under *other requests'* contexts.

Fix: always close the generator, even if the span already ended. Span behavior is unchanged. Also narrows a pre-existing `except TypeError` that could mistake an error raised by generator cleanup for the Python 3.10 `create_task` signature fallback and silently swallow it.

Alternative: #1792 builds on this and also preserves the generator's final span updates in the cancellation case. Pick whichever you prefer; happy to close the other.

<details><summary>Minimal repro (fails before, passes after)</summary>

```python
import asyncio, contextvars, gc
from langfuse import observe

request_ctx = contextvars.ContextVar("request", default="none")

@observe(name="stream")
async def stream():
    try:
        yield "a"
        yield "b"
    finally:
        print(f"finally ran under {request_ctx.get()!r}")

async def main():
    request_ctx.set("request-A")
    gen = stream()

    async def consume():
        async for _ in gen:
            await asyncio.sleep(0)

    task = asyncio.create_task(consume())
    for _ in range(4):
        await asyncio.sleep(0)
    asyncio.get_running_loop().call_soon(task.cancel)  # lands before the inner task's first step
    try:
        await task
    except asyncio.CancelledError:
        pass

    await gen.aclose()            # was a no-op: span already ended
    request_ctx.set("request-B")
    del gen; gc.collect(); await asyncio.sleep(0.05)
    # before this fix: prints "finally ran under 'request-B'"

asyncio.run(main())
```

</details>

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation update

## Verification

```bash
uv run --frozen pytest -n auto --dist worksteal tests/unit   # also run on 3.10 and 3.11
uv run --frozen ruff check .
uv run --frozen ruff format --check .
uv run --frozen mypy langfuse --no-error-summary
```

## Checklist

### DIFF
--- a/langfuse/_client/observe.py
+++ b/langfuse/_client/observe.py
@@ -735,12 +735,17 @@ class _ContextPreservedAsyncGeneratorWrapper:
 
     async def _close_generator(self) -> None:
         try:
-            await asyncio.create_task(
+            close_task = asyncio.create_task(
                 self.generator.aclose(),
                 context=self.context,
             )  # type: ignore
         except TypeError:
-            await self.context.run(asyncio.create_task, self.generator.aclose())
+            # Python 3.10: create_task has no context parameter. Only the
+            # create_task call is guarded so a TypeError raised from the
+            # generator's own cleanup is not mistaken for it.
+            close_task = self.context.run(asyncio.create_task, self.generator.aclose())
+
+        await close_task
 
     async def close(self) -> None:
         await self.aclose()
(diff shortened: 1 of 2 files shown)


DavidTraina:
superseded by https://github.com/langfuse/langfuse-python/pull/1792

PR #1792, opened 2026-07-30
fix(observe): defer span end when a cancellation never resumed the generator

DavidTraina:
## What does this PR do?

Alternative to #1791 — same bug, see there for details and a repro. Maintainers should pick one; I'll close the other.

Includes the same core fix: `aclose()`/`close()` always close the wrapped generator, even if its span already ended, plus the same `except TypeError` narrowing.

On top of that: when a cancellation ends iteration *before the generator ever resumed*, don't end the span right away. Hold the error, close the generator first — so span updates in its `finally` still land — then end the span with the held error. The span outcome is the same ERROR/"CancelledError" as today; the difference is the generator's final updates are no longer dropped.

Scope: async generators only, `CancelledError` only, and only when the generator provably never resumed. Works on Python 3.10+ (`inspect.getasyncgenstate` is 3.12+, so it falls back to the `ag_frame`/`ag_running` attributes it reads).

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation update

## Verification

```bash
uv run --frozen pytest -n auto --dist worksteal tests/unit   # also run on 3.10 and 3.11
uv run --frozen ruff check .
uv run --frozen ruff format --check .
uv run --frozen mypy langfuse --no-error-summary
```

## Checklist

### DIFF
--- a/langfuse/_client/observe.py
+++ b/langfuse/_client/observe.py
@@ -616,11 +616,7 @@ class _ContextPreservedSyncGeneratorWrapper:
 
     def close(self) -> None:
         if self._span_ended:
-            # The span can end while the generator is still suspended (e.g.
-            # an error surfaced from __next__ without resuming the generator).
-            # Still close the generator so its cleanup runs deterministically
-            # in the preserved context instead of at GC time under an
-            # arbitrary ambient context.
+            # Still close the generator so cleanup runs in the preserved context, not at GC time.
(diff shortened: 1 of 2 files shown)


hassiebp:
Thanks for your contribution, @DavidTraina !

Issue #1799, opened 2026-08-05
Langfuse.shutdown() hangs forever on _score_ingestion_queue.join() — cached LangfuseResourceManager is reused after shutdown


bagowix:
## Summary

`Langfuse.shutdown()` (and `flush()`) can block **forever** on `self._score_ingestion_queue.join()` in `LangfuseResourceManager.flush()`.

Root cause: `LangfuseResourceManager` instances are cached process-wide by `public_key` (`LangfuseResourceManager._instances`), but `shutdown()` does **not** remove the instance from the cache. Any `Langfuse` client constructed afterwards with the same `public_key` silently receives the already-shut-down manager: its ingestion consumer threads are dead, so any event enqueued after that point (e.g. `create_score()`, which checks neither `_shutdown` nor consumer liveness) is never consumed, `Queue.unfinished_tasks` never reaches 0, and the next `flush()`/`shutdown()` blocks forever on `Queue.join()`.

## Minimal reproduction

```python
import faulthandler
import threading

from langfuse import Langfuse

faulthandler.dump_traceback_later(10, exit=True)

COMMON_KWARGS = dict(
    public_key='pk-lf-repro',
    secret_key='sk-lf-repro',
    host='http://localhost:9',  # unreachable on purpose; nothing is sent in this repro
)

client_a = Langfuse(**COMMON_KWARGS)
client_b = Langfuse(**COMMON_KWARGS)

assert client_a._resources is client_b._resources  # same cached resource manager

client_a.shutdown()  # stops the shared ingestion consumer threads for good

client_b.create_score(name='quality', value=1.0)  # enqueued, but no consumer is alive anymore

print('threads alive:', [t.name for t in threading.enumerate()], flush=True)
print('score queue unfinished tasks:', client_b._resources._score_ingestion_queue.unfinished_tasks, flush=True)

client_b.shutdown()  # hangs forever on _score_ingestion_queue.join()

print('UNREACHABLE: shutdown returned', flush=True)
```

Output (the watchdog kills the process after 10 s; without it the hang is indefinite):

```
threads alive: ['MainThread', 'OtelBatchSpanRecordProcessor', 'Thread-3']
score queue unfinished tasks: 1
Timeout (0:00:10)!
...
Thread 0x00000001f38bde80 (most recent call first):
  File ".../python3.12/threading.py", line 355 in wait
  File ".../python3.12/queue.py", line 90 in join
[...]

PR #1826, opened 2026-08-20
fix(resource-manager): prevent stale clients from queueing after shutdown

trungminhdo4-glitch:
## Summary

Fixes a shutdown lifecycle race where a `LangfuseResourceManager` can accept
new score, trace, or media tasks after its background consumers have stopped.

Those tasks increment `Queue.unfinished_tasks`, but no live consumer remains to
call `task_done()`, so a later `flush()` / `shutdown()` can block indefinitely
in `Queue.join()`.

This change treats `shutdown()` as terminal for the shared resource manager:

- score/trace admission is synchronized with shutdown;
- stale clients drop asynchronous work after shutdown instead of queueing it;
- media admission is synchronized with its shutdown boundary;
- pending OTEL spans may still discover/enqueue media during `force_flush()`;
- concurrent shutdown callers wait for the active shutdown to finish;
- fork reinitialization recreates the newly introduced synchronization state.

## Why not recreate the manager automatically?

A previous approach in #1800 evicted the shut-down manager so a new same-key
client could create fresh workers.

This PR intentionally does not do that.

`LangfuseSpanProcessor` instances are registered on the shared OpenTelemetry
TracerProvider, which does not expose a normal processor-removal API.
Automatically recreating managers therefore needs a separate processor-lifetime
design to avoid accumulating registered processors.

The current SDK documentation describes `shutdown()` as application-terminal,
so this PR keeps the smaller contract: post-shutdown work cannot poison queues
or deadlock later flushes.

## Regression coverage

Five durable tests (deterministic `Event`/`Barrier` synchronization, no sleep-based timing, assert `unfinished_tasks == 0`):

- `test_1799_stale_client_score_trace_dropped_after_shutdown` — stale score/trace clients cannot enqueue after shutdown;
- `test_1799_admission_vs_shutdown_atomicity` — admission vs shutdown is atomic;
- `test_1799_concurrent_shutdown_completion` — concurrent shutdown waits for actual completion;
- `test_1799_media_admission_toctou_and_begin_shutdown` — media admission cannot cross the final shutdown boundary (`tests/unit/test_media.py`);
- `test_1799_media_during_force_flush_not_dropped_and_after_shutdown_dropped` — media discovered during OTEL `force_flush()` is still processed before media shutdown.

## Verification

- `uv run --frozen pytest tests/unit/test_resource_manager.py` — 16 passed
- `uv run --frozen pytest tests/unit/test_media.py` — 21 passed
- `uv run --frozen pytest tests/unit/test_otel.py` — 65 passed, 2 skipped
- `uv run --frozen ruff check langfuse/_client/resource_manager.py langfuse/_task_manager/media_manager.py tests/unit/test_resource_manager.py tests/unit/test_media.py` — passed
- `uv run --frozen ruff format --check langfuse/_client/resource_manager.py langfuse/_task_manager/media_manager.py tests/unit/test_resource_manager.py tests/unit/test_media.py` — passed
- `uv run --frozen mypy langfuse --no-error-summary` — passed
- `git diff --check` — passed

Broader unit run had two pre-existing `test_prompt_atexit` failures also reproducible on `main` (not introduced by this change).

Refs #1799

This PR makes resource-manager shutdown terminal and synchronizes score, trace, and media admission with consumer teardown. It also makes concurrent shutdown callers wait for completion, recreates lifecycle synchronization after forks, and adds deterministic race regression tests.

- Adds atomic shutdown checks around score and trace queue admission.
- Adds media-manager shutdown state and synchronizes queue insertion with its shutdown boundary.
- Coordinates concurrent shutdown callers through a completion event.
- Adds lifecycle and queue-draining regression coverage.
</details>

The PR should not merge until reentrant shutdown and partially admitted media traversal are handled without deadlock or dangling media references.

A callback can re-enter shutdown on the thread that must set its completion event, and an independent media traversal can cross begin_shutdown after rewriting payload values but before their upload jobs are admitted.

**Files Needing Attention:** langfuse/_client/resource_manager.py and langfuse/_task_manager/media_manager.py
</details>

```mermaid
sequenceDiagram
    participant App as Application thread
    participant RM as Resource manager
    participant OTel as Tracer provider
    participant MM as Media manager
    participant Q as Background queues
    App->>RM: shutdown()
    RM->>RM: set terminal shutdown state
    RM->>OTel: force_flush()
    OTel->>MM: discover and enqueue media
    MM->>Q: enqueue upload jobs
    RM->>Q: join queues
    RM->>MM: begin_shutdown()
    RM->>RM: stop and join consumers
    RM->>RM: set shutdown completion event
    Note over App,RM: Concurrent callers wait for completion
```
</details>

`````markdown
### Issue 1
langfuse/_client/resource_manager.py:650-655
**Reentrant shutdown self-deadlocks**

When a user-supplied span-export callback synchronously calls `shutdown()` during the initial shutdown's `force_flush()`, this branch waits on `_shutdown_event` from the same thread responsible for eventually setting it, causing shutdown to block indefinitely.

### Issue 2
langfuse/_task_manager/media_manager.py:297-302
**Shutdown creates dangling media references**

When an application thread is traversing a multi-media payload as another thread reaches `begin_shutdown()`, later `_process_media` calls return without queueing uploads while their callers still replace the payload values with media references, causing those referenced media objects never to be uploaded.

### Issue 3
tests/unit/test_resource_manager.py:457
**Test imports remain function-local**

The new regression tests place imports inside their test functions, with the same pattern repeated in the new media test and throughout this file. Move these imports to module scope to keep dependencies discoverable and follow the repository's import-placement rule.

---

`````

</details>

<sub>Reviews (1): Last reviewed commit: ["test(resource-manager): harden #1799 shu..."](https://github.com/langfuse/langfuse-python/commit/ce83473ea20970487a18d8d950eb9505e08837ae) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=55825043)</sub>

<details><summary><h4>Context used (3)</h4></summary>

- Rule used - Move imports to the top of the module instead of p... ([source](https://app.greptile.com/personal-org-4986/-/custom-context?memory=c960fc07-9928-409f-a18b-a780cbdded12))

**Learned From**
- Knowledge Base — [Client Core](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse-python/-/docs/client-core.md)
- Knowledge Base — [Task Manager: Media Upload and Score Ingestion](https://app.greptile.com/personal-org-4986/-/custom-context/knowledge-base/langfuse/langfuse-python/-/docs/task-manager.md)
</details>

### DIFF
--- a/langfuse/_client/resource_manager.py
+++ b/langfuse/_client/resource_manager.py
@@ -134,10 +134,10 @@ class LangfuseResourceManager:
         id_generator: Optional[IdGenerator] = None,
         span_exporter: Optional[SpanExporter] = None,
     ) -> "LangfuseResourceManager":
-        if public_key in cls._instances:
-            return cls._instances[public_key]
-
         with cls._lock:
+            if public_key in cls._instances:
+                return cls._instances[public_key]
+
             if public_key not in cls._instances:
                 instance = super(LangfuseResourceManager, cls).__new__(cls)
 
@@ -209,6 +209,8 @@ class LangfuseResourceManager:
         self.mask_otel_spans = mask_otel_spans
         self.environment = environment
         self._shutdown = False
+        self._shutdown_event = threading.Event()
+        self._admission_lock = threading.Lock()
 
         # Store additional client settings for get_client() to use
         self.timeout = timeout
@@ -420,6 +422,18 @@ class LangfuseResourceManager:
         # the lock is class-level state needed by the child (e.g. to create a new client)
         # even if this particular instance was already shut down.
         LangfuseResourceManager._lock = threading.RLock()
+        # Recreate shutdown event and admission lock for child process
+        if hasattr(self, "_shutdown_event"):
+            self._shutdown_event = threading.Event()
+            if self._shutdown:
+                self._shutdown_event.set()
+        if hasattr(self, "_admission_lock"):
+            self._admission_lock = threading.Lock()
+        # Recreate media manager state lock if held at fork
+        if hasattr(self, "_media_manager") and hasattr(
+            self._media_manager, "_state_lock"
+        ):
+            self._media_manager._state_lock = threading.Lock()
 
         if self._shutdown:
             return
@@ -508,10 +522,16 @@ class LangfuseResourceManager:
             )
 
             if should_sample:
-                langfuse_logger.debug(
-                    f"Score: Enqueuing event type={event['type']} for trace_id={event['body'].trace_id} name={event['body'].name} value={event['body'].value}"
-                )
-                self._score_ingestion_queue.put(event, block=False)
+                with self._admission_lock:
+                    if self._shutdown:
+                        langfuse_logger.warning(
+                            "Score: Dropping event because the Langfuse client has already been shut down."
+                        )
+                        return
+                    langfuse_logger.debug(
+                        f"Score: Enqueuing event type={event['type']} for trace_id={event['body'].trace_id} name={event['body'].name} value={event['body'].value}"
+                    )
+                    self._score_ingestion_queue.put(event, block=False)
 
         except Full:
             langfuse_logger.warning(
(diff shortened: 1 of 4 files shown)


trungminhdo4-glitch:
Rebase note: this PR is CONFLICTING on langfuse/_client/resource_manager.py (2 hunks, score + trace paths). Upstream restyled the debug logging to %-format in the same lines where this PR adds the admission-lock shutdown gate. I resolved it locally and verified the full rebase replays cleanly onto current main (both commits, py_compile clean): keep the admission-lock + shutdown-drop gate, adopt the upstream %-style debug call inside it, i.e. gate first, then the debug log, then queue.put(block=False).

Blocker on my side: git-protocol access to all my forks currently 404s/401s (REST API works fine, e.g. this comment), so I cannot push the rebased branch right now - investigating, will push the moment transport works. Happy to rework the resolution if you prefer the f-string style kept.

PR #1845, opened 2026-08-28
perf(span-processor): skip span_formatter when debug logging is off

### LINKED ISSUE #15339: perf: span_processor.on_end evaluates span_formatter(span) eagerly via f-string regardless of log level
### Summary

`LangfuseSpanProcessor.on_end` in `langfuse/_client/span_processor.py` evaluates `span_formatter(span)` eagerly inside an f-string passed to `langfuse_logger.debug(...)`. Because Python f-strings are evaluated *before* being handed to the logger, the full span serialization runs on **every** span end, regardless of the effective log level. This is a measurable per-span CPU cost on the thread that ends the span (in async apps, the event loop thread), and cannot be mitigated by raising the log level.

### Environment

- Langfuse Python SDK: `langfuse-python` 4.14.0 (also confirmed still present on `main` as of 2026-07-23)
- Python 3.12
- OS: Linux (containerized, FastAPI + asyncio)

### Reproducible example

The offending code (current `main`, `langfuse/_client/span_processor.py`, `LangfuseSpanProcessor.on_end`):

```python
langfuse_logger.debug(
    f"Trace: Processing span name='{span._name}' | Full details:\n{span_formatter(span)}"
)
```

`span_formatter(span)` performs a full `json.dumps` of the span (attributes, events, links, resource, instrumentationScope). Profiling a decorated async handler returning ~1000 spans/s shows this single call at ~0.27 ms/span, i.e. ~27% of one core purely on formatting a debug message that is never emitted (logger at WARNING).

Crucially, setting `LANGFUSE_LOG_LEVEL=WARNING` (or `debug=False`) does **not** help: the f-string is constructed before `logger.debug` is called, so the formatting cost is paid unconditionally.

### Root cause

Python logging defers %-formatting of *arguments* (via `logger.debug(msg, *args)`), but it cannot defer evaluation of expressions already embedded in an f-string. The f-string is built at call time, so `span_formatter(span)` runs even when the DEBUG level is disabled. This is the well-known f-string-in-logging anti-pattern (https://stackoverflow.com/questions/54367975).

### Impact

- Per-span overhead on the calling thread (event loop in async apps) → contributes to GIL pressure under high span throughput.
- In our deployment this combined with high-frequency spans to saturate CPU and cause service restarts. Even in isolation it is a non-trivial tax proportional to span throughput.
- Not mitigated by log level configuration, which is counter-intuitive for users trying to reduce SDK overhead.

### Suggested fix

Guard the formatting with an enabled check (cheapest, no API change):

```python
if langfuse_logger.isEnabledFor(logging.DEBUG):
    langfuse_logger.debug(
        f"Trace: Processing span name='{span._name}' | Full details:\n{span_formatter(span)}"
    )
```

or use lazy %-style args (note: `span_formatter(span)` would still evaluate as an arg, so the `isEnabledFor` guard is preferred; alternatively defer via a small lazy wrapper).

### Additional context

- A related but **different** closed issue is #4863 (serializer `print(obj.__dict__)` dumping object graphs). That was fixed by removing the `print`; the `on_end` f-string here was not addressed.
- The same f-string eager-eval pattern also appears in `on_start` (`f"Propagated {len(...)} attributes..."`) and several `on_end` debug branches; worth auditing together.

DavidTraina:
## What does this PR do?

Fixes langfuse/langfuse#15339

`LangfuseSpanProcessor.on_end` built its debug message with an f-string:

```python
langfuse_logger.debug(
    f"Trace: Processing span name='{span._name}' | Full details:\n{span_formatter(span)}"
)
```

f-strings are evaluated before `debug()` is called, so `span_formatter(span)` -- a `json.dumps(indent=2)` of the span's attributes, events, links, resource and scope -- ran on every span end even though the `langfuse` logger defaults to `WARNING` and the string was then discarded. `on_end` runs on the thread that ends the span, the event loop in async apps, so this was a per-span CPU cost proportional to span throughput that no log-level setting could switch off.

Measured locally, `span_formatter` costs 28 us on a small span and 167 us on a 100 KB one; the linked issue reports ~0.27 ms/span. The `isEnabledFor` check costs 0.02 us.

This guards the call so the serialization only runs when the record would be emitted, and passes the message as lazy `%`-style args. Output at DEBUG is unchanged.

Adds a parametrized unit test that drives a real span through the processor and asserts `span_formatter` is not called at `WARNING` and is called once at `DEBUG`. The `WARNING` case fails on `main`.

Related: #1846 converts the SDK's remaining f-string log calls to lazy formatting and enables ruff `G004`. It touches this same call to make it lazy but does not add the guard, so whichever merges second needs a trivial rebase here.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor
- [ ] Documentation update

## Verification

```bash
uv run --frozen ruff check .                                        # All checks passed!
uv run --frozen ruff format --check .                              # no drift in changed files
uv run --frozen mypy langfuse --no-error-summary                    # exit 0
uv run --frozen pytest tests/unit/test_span_processor.py            # 4 passed
uv run --frozen pytest -n auto --dist worksteal tests/unit          # 688 passed, 2 skipped
```

e2e / live_provider were not run: unit-testable change, no network behavior touched.

## Checklist

- [x] I updated docs, examples, or `.env.template` if needed. (n/a, internal-only change)

### DIFF
--- a/langfuse/_client/span_processor.py
+++ b/langfuse/_client/span_processor.py
@@ -12,6 +12,7 @@ Key features:
 """
 
 import base64
+import logging
 import os
 import threading
 from typing import Callable, Dict, List, Optional, cast
@@ -218,9 +219,13 @@ class LangfuseSpanProcessor(BatchSpanProcessor):
                 )
                 return
 
-            langfuse_logger.debug(
-                f"Trace: Processing span name='{span._name}' | Full details:\n{span_formatter(span)}"
-            )
+            # span_formatter serializes the full span; skip it unless DEBUG is on
+            if langfuse_logger.isEnabledFor(logging.DEBUG):
+                langfuse_logger.debug(
+                    "Trace: Processing span name='%s' | Full details:\n%s",
+                    span.name,
+                    span_formatter(span),
+                )
 
             super().on_end(span)
         finally:
--- a/tests/unit/test_span_processor.py
+++ b/tests/unit/test_span_processor.py
@@ -1,8 +1,12 @@
+import logging
 from typing import Sequence
+from unittest.mock import patch
 
-from opentelemetry.sdk.trace import ReadableSpan
+import pytest
+from opentelemetry.sdk.trace import ReadableSpan, TracerProvider
 from opentelemetry.sdk.trace.export import SpanExporter, SpanExportResult
 
+import langfuse._client.span_processor as span_processor_module
 from langfuse._client.environment_variables import (
     LANGFUSE_FLUSH_AT,
     LANGFUSE_FLUSH_INTERVAL,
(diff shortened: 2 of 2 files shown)


DavidTraina:
Related: #1846 converts the remaining 123 f-string logging calls in `langfuse/` to lazy `%`-style formatting. It deliberately leaves the `span_formatter` call in `on_end` untouched so it does not overlap with this PR — the two can merge in either order.

